PR #6, opened 2021-04-04
xrfc: xdstp:// xRFC (TP1)

htuch:
This ports
https://docs.google.com/document/d/1zZav-IYxMO0mP2A7y5XHBa9v0eXyirw1CxrLLSUqWR8/edit#
to xRFC format and updates to match changes that have occurred during
Envoy implementation.

This is xRFC number TP1.

Signed-off-by: Harvey Tuch <htuch@google.com>


htuch:
CC @cncf/xds-wg

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+Context parameters in URNs presented by the client to the server will be dervied
+by the following process. Using an example of a URL
+`xdstp://some-authority/some.type/foo?xds.bar=baz`:
+
+1. Static `Node`-derived context parameters are first added. These are prefixed with
+   `xds.node.`. The set of `Node`-derived context parameters is specified in the
+   bootstrap on a per-resource type basis. The key reflects the `Node` proto3
+   field structure, e.g. `xds.node.locality.sub_zone=some_sub_zone`,
+   `xds.node.metadata.bar="a"`, `xds.node.user_agent_version=1.2.3`.
```
howardjohn, line 280: I think metadata can handle arbitrary structs. How will these be encoded? In particular lists, lists of structs, and `"5"` vs `5` seem like tricky cases.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+5. Finally per-resource type well-known attributes are added, e.g. an on-demand
+   listener load might have `{xds.resource.vip: 96.54.3.1}`. These attributes
+   are generated by code, e.g. an on-demand LDS implementation will first detect
+   the VIP, halt the request, ask for the `xds.resource.vip` from the control
+   plane, and then resume the request after discovery is completed. All these
+   attributes will be `xdstp.resource.` prefixed.
+
+An example computed URN following the above example is
+`xdstp://some-authority/some.type/foo?xds.node.metadata.foo=bar&xds.shard_id=1234&xds.bar=baz&xds.client_feature.lb.least_loaded=false&xds.resource.vip=96.54.3.1`.
```
howardjohn, line 305: This will probably be massive for some cases - well over 2048 bytes which seems to be the limit for some libraries. But maybe that won't be an issue here since we are not HTTP. I am not familiar with enough URI parsing libraries to know if any of them will complain or if its just internet explorer which is irrelevant here.

htuch: If names grow this large, we can encounter wire performance problems. We've spent a bit of time thinking on how to optimize this on the transport, e.g. factor out common context parameters. The thing to keep in mind is that URIs are a logical view of naming as well as concrete syntax; we can always optimize internally and on the wire (with protobuf) to avoid explicit string representation when things become unwieldy, but ideally applications limit the scope and size of context parameters.

proposals/TP1-xds-transport-next.md
htuch: It's a good question; you can see the reference implementation in Envoy for this https://github.com/envoyproxy/envoy/blob/a45ce6578caa682e4556a671261772730dcdfd65/source/common/config/xds_context_params.cc#L79. 

Should we specify in more detail what it does here or have some other way of sharing reference implementation (e.g. move the code to the `cncf/xds` repository)?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+    InlineEntry inline_entry = 2;
+  }
+}
+```
+
+A simple example of a listener collection fetch sequence:
+
+1. Client requests `xdstp://some-authority/envoy.config.listener.v3.ListenerCollection/foo.`
+2. Server responds with a list of resources: `[locator:
+   xdstp://some-authority/envoy.config.listener.v3.Listener/bar, locator:
```
fuqianggao, line 366: if the list request has some context params, does the list response need to include those as well? Same question for glob.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+on individual resources but the resource list itself can become a bottleneck for
+frequently updated large collections of resources, e.g. with 1M endpoints updating
+every 10s.  For this situation, we introduce *glob
+collections*, which allow the `<T>Collection` directory to be elided when
+resources in a collection share a directory structure and a `/*` component is appended to
+the resource path. Continuing the previous example, this now looks like:
+
+1. Client requests
+   `xdstp://some-authority/envoy.config.listener.v3.Listener/foo/*.`
+2. Server responds with resources
```
fuqianggao, line 395: Are we going to support inlining resource for glob?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+State-of-the-world (SotW) `Discovery{Request,Response}` are not supported with
+the new naming scheme and are effectively deprecated for servers that make use
+of the scheme. The use of inlined entries in list collections allows simple
+management servers that do not wish to perform state tracking to perform
+state-of-the-world updates. This is achieved by sending the entire state, e.g.
+every listener inlined in a `ListenerCollection`.
+
+In both the xDS transport and resources, existing fields for managing resource
+names will continue to be used. A client and server will detect `xdstp://`
```
fuqianggao, line 485: do we consider to add a client feature capability explicitly for client to tell management server if it supports the xdstp naming scheme?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+bootstrap.
+
+Client EDS `DeltaDiscoveryRequest` sends to xDS management server for `some-cloud-authority`:
+
+
+```
+resource_names_subscribe:
+- xdstp://some-cloud-authority/envoy.config.endpoint.v3.ClusterLoadAssignment/foo
+```
+
```
fuqianggao, line 690: can you expand on this. What does this mean?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+   plane, and then resume the request after discovery is completed. All these
+   attributes will be `xdstp.resource.` prefixed.
+
+An example computed URN following the above example is
+`xdstp://some-authority/some.type/foo?xds.node.metadata.foo=bar&xds.shard_id=1234&xds.bar=baz&xds.client_feature.lb.least_loaded=false&xds.resource.vip=96.54.3.1`.
+
+This proposal reserves the `@` prefix for context parameter values for future
+URI context parameter enhancements.
+
+### Collections
```
fuqianggao, line 310: Do you want to mention which resource type will have list/glob support? e.g. lds, cds.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+* [URL](https://github.com/cncf/xds/blob/main/xds/core/v3/resource_locator.proto) definition
+* [URN](https://github.com/cncf/xds/blob/main/xds/core/v3/resource_name.proto) definition
+
+#### Normalization
+
+Two `xdstp://` URNs are considered equivalent if they match component-wise
+modulo context parameter ordering.
+
+## Context parameters
```
norafang52, line 270: How are we going to handle duplicate context params from different process? Say one from URL has same key with the one from dynamic context param added by client.

mattklein123 (commented):
Thanks for porting this over. A few comments from a fresh read through.

proposals/TP1-xds-transport-next.md
htuch: https://docs.google.com/document/d/1zZav-IYxMO0mP2A7y5XHBa9v0eXyirw1CxrLLSUqWR8/edit#heading=h.xbjx2ihsbbaa covers the HTTP transport. I have not included this yet, since I'm trying to keep the Markdown to a fairly limited implementable-in-the-near-term scope. We can port that stuff across when we start work on the HTTP support.

proposals/TP1-xds-transport-next.md
htuch: No, since there is nothing to inline into. For a list collection, you can inline into the directory message.

proposals/TP1-xds-transport-next.md
htuch: The client basically announces this through its use of `xdstp://` names. It probably would only be useful to do this in `Node` if we do it, since if yo're using `xdstp://` you don't need to announce you're using it :D Would this be useful to control plane implementors from your perspective?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+1. Static `Node`-derived context parameters are first added. These are prefixed with
+   `xds.node.`. The set of `Node`-derived context parameters is specified in the
+   bootstrap on a per-resource type basis. The key reflects the `Node` proto3
+   field structure, e.g. `xds.node.locality.sub_zone=some_sub_zone`,
+   `xds.node.metadata.bar="a"`, `xds.node.user_agent_version=1.2.3`.
+
+2. Dynamic context parameters are added next. These are free form and at the
+   discretion of the client. They can provide client-specific fields such as
+   `xds.shard_id=1234` and potentially change through the lifetime of an xDS
+   stream.
```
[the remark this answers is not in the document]
htuch: The context parameter changes in the client. This then causes the client to ask for a new request from the management server. The management server only observes a new request with an `xdstp://` URL with different context parameters, caching still continues to work at the management server, since the request now refers to a different resource.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+4. [Client feature
+   capability](https://www.envoyproxy.io/docs/envoy/latest/api/client_features)
+   context parameters are then added, e.g. clusters at the client might have
+   `{xds.client_feature.lb.least_loaded: true}`. Client feature capabilities
+   will be required to be declared with respect to a specific resource type URL.
+   All these attributes will be `xds.client_feature.` prefixed.
+
+5. Finally per-resource type well-known attributes are added, e.g. an on-demand
+   listener load might have `{xds.resource.vip: 96.54.3.1}`. These attributes
```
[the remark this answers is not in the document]
htuch: Correct.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
mattklein123: Yes that makes sense. Rephrase? It sounds like this can change mid-stream.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+   `xds.shard_id=1234` and potentially change through the lifetime of an xDS
+   stream.
+
+3. Context parameters from the URL are added next, in the above example
+   `xds.bar=baz`.
+
+4. [Client feature
+   capability](https://www.envoyproxy.io/docs/envoy/latest/api/client_features)
+   context parameters are then added, e.g. clusters at the client might have
+   `{xds.client_feature.lb.least_loaded: true}`. Client feature capabilities
```
markdroth, line 293: I think we had also said that these might be a little bit viral, because some resources can be sent as part of another resource.  For example, the RouteConfig may be inlined into the Listener resource or it may be obtained separately, and the client doesn't know when it sends the LDS request whether the response will include the RouteConfig.  This means that if the client has a feature that affects the RouteConfig, it needs to include that in all LDS or RDS requests it sends.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+## Proposal
+
+The proposal below is fully backwards compatible with the existing xDS transport
+protocol and does not require any changes to clients or servers that do not wish
+to opt into the new naming scheme.
+
+### URI-based xDS resource names
+
+The following URI schema is proposed for xDS resource names:
```
markdroth, line 226: Might be a good idea to explicitly say that this is an RFC-3986-compliant schema, with a link to https://tools.ietf.org/html/rfc3986.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+Multiple directives are comma separated in the fragment component of the URI.
+
+An example application of the above URI scheme is:
+
+`xdstp://some.control.plane/envoy.config.route.v3.RouteConfiguration/foo/bar?shard_id=1234&direction=inbound`
+
+We distinguish between two forms of `xdstp://` URIs:
+
+* Uniform Resource Locators (URLs) instructing a client on how to locate a given
```
markdroth, line 251: How about also saying that URLs can include globs, while URNs cannot?

proposals/TP1-xds-transport-next.md
markdroth: From the perspective of the transport protocol, I don't think there's any limitation here -- there can be collections for any resource type for which that proves useful.

In practice, LDS and CDS are the two main ones today, but we'll also use glob collections for LEDS, and there may be others in the future.  So I don't think we can have an exhaustive list here.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+```
+
+where `CollectionEntry` is defined:
+
+
+```
+message CollectionEntry {
+  // Inlined resource entry.
+  message InlineEntry {
+    // Optional name to describe the inlined resource. Resource names must
```
markdroth, line 336: We should probably note that this must be unique within the list collection.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+    // The resource payload, including type URL.
+    google.protobuf.Any resource = 3;
+  }
+
+  oneof resource_specifier {
+    option (validate.required) = true;
+
+    // A resource locator describing how the member resource is to be located.
+    ResourceLocator locator = 1;
```
markdroth, line 354: Should this just be a `string`, considering that we're using the string form everywhere else?  (As mentioned above, I know that we've been discussing introducing this structured form for wire-efficiency reasons, but I assume that will be covered in a subsequent xRFC.)

proposals/TP1-xds-transport-next.md
markdroth: From the perspective of the transport protocol, a list collection is a single opaque resource, just like any other resource.  The fact that this particular resource embeds other resources inside of it is visible at the data model layer, but the transport protocol layer doesn't know or care.  So the context paramerter rules here apply exactly the same way as for any other resource: the exact name requested by the client must be returned by the server, including all context params.

Glob collections are basically the same thing.  The transport protocol knows about expanding the glob for the "id" part of the URL, so it can match multiple URNs returned by the server to a single URL, but all of the other parts of the URL (authority, resource type, and context params) must exactly match.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+1. Client requests
+   `xdstp://some-authority/envoy.config.listener.v3.Listener/foo/*.`
+2. Server responds with resources
+   [`xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar,
+   xdstp://some-authority/envoy.config.listener.v3.Listener/foo/baz].`
+
+Since each resource returned is subject to independent update via delta xDS and
+there is no explicit collection directory to update, glob collections are highly
+scalable.
```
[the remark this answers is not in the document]
markdroth: Consider the LEDS case.  We want the server to be able to do two things efficiently:

1. Send updates for individual endpoints that have changed without resending other endpoints that have not changed.
2. Dynamically change the *set* of endpoints for a given cluster -- i.e., to add and remove endpoints from the set without resending the whole set.

In list collections, there is one resource defining the set of resources and another resource defining the individual resources (assuming non-inlined entries in the collection).  So with a list collection, it's possible to do (1), but it's not possible to do (2) -- adding or removing an endpoint from the set would require resending the entire list collection resource.  Glob collections allow us to do (2), because we can simply send an add or a delete for individual endpoints that we want to add or remove from the set.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+An `xdstp://` URI specifies a logic authority and it’s necessary to map to some physical 
+network address during the resource fetch process. In the existing xDS configuration
+fetch configuration, this capability is provided by a `ConfigSource` delivered alongside a
+resource name in the xDS resource graph.
+
+This proposal adds to `ConfigSource` the ability to specify the
+authorities for which the `ConfigSource` may be used:
+
+```
+message Authority {
```
markdroth, line 444: For gRPC, I think we're going to want to structure this differently.  We're going to have a map of authority name to the configuration for that authority.  The config for a given authority will include the following:

- An ordered list of (the equivalent of) `ConfigSource`s that indicate which server(s) can provide data for that authority.  The client will prefer servers in the order listed here, which provides a convenient way of configuring failover.
- Node context params will be specified here, since these are likely to be different for different authorities.
- (Not part of this design, but in the follow-up xRFC I plan to write for dynamic context parameter negotiation, this will include setting the name of the DCP resource that will be used for this authority.)

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+  …  extant ConfigSource … 
+}
+```
+
+When the `ConfigSource` adjacent to a resource name matches the URI authority,
+the `ConfigSource` will continue to be used.
+
+When there is no match, the xDS client will fallback to the bootstrap to map
+from authority to `ConfigSource.
```
markdroth, line 461: Formatting nit: Missing a back-tick here.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+used in `xdstp://` authority resolution, an `ConfigSource` will act as an ADS
+transport if configured with this new enum value. Instead of pointing to a
+single shared ADS `ConfigSource` declared in the bootstrap, any resource
+authority (regardless of type) that maps to a `ConfigSource` with
+`AGGREGATED_DELTA_GRPC` will be multiplexed on the xDS stream specified by the
+`ConfigSource`.
+
+### Discovery request and responses
+
+State-of-the-world (SotW) `Discovery{Request,Response}` are not supported with
```
markdroth, line 477: Now that we're using strings instead of structured protos for URIs on the wire, I don't think there's actually any reason to not support xdstp names in SotW.  I think the only thing that doesn't work in SotW is glob collections, because there's no  way for the server to delete a resource from the collection.

proposals/TP1-xds-transport-next.md
markdroth: Yeah, I don't see how a client capability will help here.  The client has to send the resource name it wants in the same message as it sends its client capabilities, so by the time the server sees the capability, it's already too late.

I think the way that people will use this is simply to not configure clients to use xdstp names unless they're talking to a control plane that supports those names.  After all, just like an HTTP server, it's the xDS server that determines which names match to which contents; the client has to ask for the name that the server is using for whatever content it wants.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+In both the xDS transport and resources, existing fields for managing resource
+names will continue to be used. A client and server will detect `xdstp://`
+semantics by examining the prefix of a resource string.
+
+Clients are expected to have knowledge ahead of time (via mechanisms not part of
+this proposal, e.g. support documentation) on whether the servers specified in
+the bootstrap supports the new URI conventions.
+
+The server may issue redirects by including an `xds::core::v3::ResourceLocator`
```
markdroth, line 492: I wonder if this paragraph should be in its own section.  The redirect mechanism is important, but it's a little buried in this section, so folks just scanning the doc at a high level might not see it.

markdroth (commented):
Thanks for doing this!

proposals/TP1-xds-transport-next.md
htuch: Updated, hopefully this captures the rendering as described in code.

proposals/TP1-xds-transport-next.md
htuch: Yep. Let me add this to the examples to be clear.

proposals/TP1-xds-transport-next.md
htuch: Which bit is unclear? Happy to expand on anything here, just need to figure out what is missing :)

proposals/TP1-xds-transport-next.md
htuch: Yep, true, I added LDS, CDS and LEDS as non-exhaustive examples.

proposals/TP1-xds-transport-next.md
htuch: @markdroth did we end up with any rules around how URL parameters (i.e. free form) are supposed to work? I.e. "they can never starts with xds." or "they can optionally start with xds. if deliberately overriding other context parameters"?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+by the following process. Using an example of a URL
+`xdstp://some-authority/some.type/foo?xds.bar=baz`:
+
+1. Static `Node`-derived context parameters are first added. These are prefixed with
+   `xds.node.`. The set of `Node`-derived context parameters is specified in the
+   bootstrap on a per-resource type basis. The key reflects the `Node` proto3
+   field structure, e.g. `xds.node.locality.sub_zone=some_sub_zone`,
+   `xds.node.metadata.bar="a"`, `xds.node.user_agent_version=1.2.3`.
+
+2. Dynamic context parameters are added next. These are free form and at the
```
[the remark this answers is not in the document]
htuch: I think we need to be clear on the relative ordering, since they are somehow combined with URL-derived context parameters and we need to be clear which way the override relationship works. Since we have dynamic parameters that can come from the binary, as well as the transport negotiation that you are working on, it probably makes sense to touch on them here.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: Updated, LMK WDYT.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: The next section, `Use cases`, goes into this, but I've updated with an example clarifying the scalability argument.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+
+We distinguish between two forms of `xdstp://` URIs:
+
+* Uniform Resource Locators (URLs) instructing a client on how to locate a given
+  xDS resource. URLs may include processing directives.
+
+* Uniform Resource Names (URNs) referring to a specific xDS resource. URNs do not
+  include processing directives. `xdstp://` URNs provide self-contained cache keys.
+
+#### Proto3 expression
```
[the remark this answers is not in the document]
htuch: I could go either way on this one; there is some useful clarifying detail in the comments in the repository. I almost feel it's cleaner to keep that as the canonical data model and have this live as literate code. But I could also be persuaded that the xRFC should be standalone and we should fold any detail in those protos into the doc.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+Two `xdstp://` URNs are considered equivalent if they match component-wise
+modulo context parameter ordering.
+
+## Context parameters
+
+Context parameters in URNs presented by the client to the server will be dervied
+by the following process. Using an example of a URL
+`xdstp://some-authority/some.type/foo?xds.bar=baz`:
+
+1. Static `Node`-derived context parameters are first added. These are prefixed with
```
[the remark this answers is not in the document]
htuch: Yes, but see my comments on when there is a decision to deliberately override or the relative ordering of dynamic/static URL params.

proposals/TP1-xds-transport-next.md
htuch: Thanks for reminding me; I've updated the text to explain this.

proposals/TP1-xds-transport-next.md
htuch: True, but I rather think about this additively; land the delta spec and add SotW as needed.

proposals/TP1-xds-transport-next.md
htuch: While I've updated it, I'm actually wondering if this is getting a bit impractical; if you have something like filters, which can be literally _anything_, this seems a lot of client features to be sending. E.g. We need to send client features for the Dubbo filter references, as well as those for a RouteConfiguration embedded in HCM. 

Relying on the user agent (name+version) and having some well known map from that to client features starts to sound attractive.

proposals/TP1-xds-transport-next.md
htuch: What about the `ConfigSource` that is not in the bootstrap but already sitting next to the resource in question, e.g. https://github.com/envoyproxy/envoy/blob/4ecf3f7efed97529dc064223f3697a8d4d1101b9/api/envoy/config/filter/network/http_connection_manager/v2/http_connection_manager.proto#L528. Should there be some way to allow this to be optionally used as the `ConfigSource` for the authority as a bootstrap override? It seems weird to give up on allowing dynamic config source completely when we have this flexibility today.

proposals/TP1-xds-transport-next.md
```diff
@@ -398,7 +425,18 @@ the resource path. Continuing the previous example, this now looks like:
...
+collections requires that the collection directory (now containing 10001
+resource references) to be sent from server to client, as well as the additional
+new resource.
+
+As with list collections, context parameters in the request must be matched in
+responses. A request for
+`xdstp://some-authority/envoy.config.listener.v3.Listener/foo/*?some=thing` will
+match a resource
+`xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar?some=thing`
+but not `xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar`.
```
mattklein123, line 439: So this makes sense to me once @markdroth explained that globs support delta (add/remove), but pedantically the text above and example do not make that clear. I understand you have it in examples below, but is there a way that you can make this section more clear that glob supports delta add/remove?

proposals/TP1-xds-transport-next.md
fuqianggao: How is the fallback to 'some-onprem-authority' configured?

proposals/TP1-xds-transport-next.md
norafang52: I think it will be clear to have some rules for context parameters from URL.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1068 @@
...
+1. Client requests
+   `xdstp://some-authority/envoy.config.listener.v3.Listener/foo/*.`
+2. Server responds with resources
+   [`xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar,
+   xdstp://some-authority/envoy.config.listener.v3.Listener/foo/baz].`
+
+Since each resource returned is subject to independent update via delta xDS and
+there is no explicit collection directory to update, glob collections are highly
+scalable. For example, consider a collection of 10k resources and the addition
+of a single resource. Glob collections will send the single additional resource. List
```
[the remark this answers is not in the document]
htuch: Yes; this section starts of talking about delta xDS. I can reiterate, the assumption of delta seems to be causing multiple folks to question the meaning.

proposals/TP1-xds-transport-next.md
htuch: Friendly ping on this one @markdroth. I'm happy to add a sketch of a bootstrap and authority mapping, but it would be good to describe how this interoperates with existing `ConfigSource` that sit next to resource names.

proposals/TP1-xds-transport-next.md
htuch: The presence of the `#alt` directive above describes the fallback relationship. It's also mentioned that there is a mapping from `some-onprem-authority` to `ConfigSource` in the bootstrap.

proposals/TP1-xds-transport-next.md
htuch: Sure. @markdroth can you comment on https://github.com/cncf/xds/pull/6#discussion_r611273789?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1071 @@
...
+* `resource type` is the xDS resource type URL, e.g. `envoy.config.route.v3.RouteConfiguration`
+* `id/*` is the remaining path component of the URI and is fully opaque; naming
+  is at the discretion of the control plane(s).
+* `context parameters` are the URI query parameters and express contextual
+  information for selecting resource variants, for example `shard_id=1234&direction=inbound`.
+* A `processing directive` provides additional information to the client on how
+  the URI is to be interpreted. Two supported directives are:
+  * `alt=<xdstp:// URI>` expressing a fallback URI.
+  * `entry=<resource name>` providing an anchor referring to an inlined resource
+    for [list collections](#list). Resource names must be of the form `[a-zA-Z0-9_-\./]+`.
```
adisuissa, line 242: Can more than one _processing directive_ appear in a locator?
If so, can the same directive appear more than once?

adisuissa (commented):
Overall looks very good!
Left a couple of questions and minor nits.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1071 @@
...
+on individual resources but the resource list itself can become a bottleneck for
+frequently updated large collections of resources, e.g. with 1M endpoints updating
+every 10s.  For this situation, we introduce *glob
+collections*, which allow the `<T>Collection` directory to be elided when
+resources in a collection share a directory structure and a `/*` component is appended to
+the resource path. Continuing the previous example, this now looks like:
+
+1. Client requests
+   `xdstp://some-authority/envoy.config.listener.v3.Listener/foo/*.`
+2. Server responds with resources
```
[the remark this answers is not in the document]
htuch: This is a good question. I think we have a choice between not responding and responding with a sentinel value. I think not responding might delay warming, which will matter for something like LEDS. How about we allow a `Resource` to be returned with the `Any` field not set and the name of the glob collection as a sentinel here?

@markdroth thoughts?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1071 @@
...
+Client EDS `DeltaDiscoveryRequest` sends to xDS management server for `some-cloud-authority`:
+
+
+```
+resource_names_subscribe:
+- xdstp://some-cloud-authority/envoy.config.endpoint.v3.ClusterLoadAssignment/foo
+```
+
+Under normal conditions, the requested resource is returned as above. If the
+config source for `some-cloud-authority` is unreachable, then the client issues
```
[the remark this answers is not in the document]
htuch: This is a good question; I'm tempted to punt as we're not implementing this right now and it gets complicated. What about leaving it as a client specific how failover and restoration of original resource fetch is decided?

@markdroth also interested in your take on this one.

proposals/TP1-xds-transport-next.md
htuch: Yes on > 1 processing directive, see next sentence. I've clarified on the same directive multiple times.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
adisuissa: IMO explicit is better as it allows distinguishing between timeout and non-existence, and may simplify things later with alternatives (in case of a timeout the alternative will be used, but it probably isn't the desired behavior).

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1028 @@
...
+    "@type": type.googleapis.com/envoy.config.listeners.v3.ListenerCollection
+    - locator: xdstp://some-authority/envoy.config.listeners.v3.Listener/bar
+    - locator: xdstp://some-authority/envoy.config.listeners.v3.Listener/baz
+```
+
+
+The `bar` and `baz` resources are then fetched as per [singleton resource
+requests](#singleton-resource-request).
+
+#### List collections with inlining
```
norafang52: How would server know whether to include inline resources or not?

htuch: I added a comment below, but basically this is at the discretion of the server and it decides based on criteria such as reducing round-trips.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: I added a note stating this is client specific.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: I've added a proposal to use sentinel empty resource,

htuch:
I've updated based on feedback. @fuqianggao @norafang52 @markdroth @adisuissa PTAL.

proposals/TP1-xds-transport-next.md
markdroth: I think where we wound up is that the ad-hoc context params can't start with `xds.`, so there will basically never be any conflicts.

proposals/TP1-xds-transport-next.md
markdroth: gRPC does not currently support `ConfigSources` that sit next to resource names (which I refer to as "dynamic `ConfigSource`s") that point at arbitrary remote servers, because there are concerns about trusting the management server to tell the client what credentials to use to talk to a remote server.  Currently, our bootstrap file specifies a single ADS server from which we get all of our data, and we require dynamic `ConfigSource`s to populate the `ads` field, so that they always specify resources from the same server.  As part of implementing federation in gRPC, we will change our bootstrap file to support the authority map I described above, and then servers will be able to tell clients to get data from a different server via the authority in the resource URI, as long as that authority is already present in the client's bootstrap file.  But we still don't plan to allow servers to send arbitrary dynamic `ConfigSource`s to clients.  (The only change we will make is that we will start allowing the `self` field, and we will treat `ads` the same as `self`  for backward compatibility, which will take us in the same direction as envoyproxy/envoy#13951.)  So for gRPC, this is not an issue.

That having been said, if Envoy is going to continue to support these dynamic `ConfigSource`s, it seems to me that you can continue to support that regardless of how the per-authority `ConfigSource`s are laid out in your bootstrap file.  I think your original plan was that if there is a dynamic `ConfigSource`, it will be used, and if not, you will look up the authority from the resource name in the bootstrap file to see if there's a `ConfigSource` present for that authority.  I think you can stick with that approach, regardless of what exact format you use for the `ConfigSource`s in the bootstrap file.

I think the main advantages of having an authority map in the boostrap file are that (a) it provides a way to specify the priority order of the list of servers to use for a given authority and (b) it provides a place to put other per-authority configuration, such as cache key constraint configuration.

proposals/TP1-xds-transport-next.md
markdroth: I'm okay with that, but I do think that we'll need another xRFC for the new naming scheme in SotW in the near future, because gRPC is planning to implement that.  And that xRFC will be very short; it will basically say "everything from the original xRFC except for glob collections works for StoW too".

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
markdroth: I think the client should always be trying to contact the original server, with appropriate backoff delay.  Whenever the original server recovers enough to return a result, then the client should switch back to data from that server.

That having been said, I think I would be in favor of removing this from this xRFC, because we don't plan to implement this part in either gRPC or in Envoy in the near future.  I think we can put together a separate xRFC for this once we actually have a need for it.

markdroth (commented):
Sorry it took me so long to re-review this!  Please let me know if you have any questions about any of this.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1082 @@
+xdstp:// structured resource naming, caching and federation support.
```
markdroth, line 1: Minor nit: Please add "TP1:" at the start of this line, so the xRFC number is clear to readers.

Might also be a good idea to add the xRFC number to the PR description, too.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1082 @@
...
+in their naming.
+
+#### Glob
+
+Having explicit collection resources allows delta xDS to provide delta updates
+on individual resources but the resource list itself can become a bottleneck for
+frequently updated large collections of resources, e.g. with 1M endpoints updating
+every 10s.  For this situation, we introduce *glob
+collections*, which allow the `<T>Collection` directory to be elided when
+resources in a collection share a directory structure and a `/*` component is appended to
```
norafang52, line 418: Based on this definition,  a glob collection might ask for resources under sub-directory as well right?
Eg. xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar/*

And request with multiple glob resources is also possible?
Eg. ["xdstp://some-authority/envoy.config.listener.v3.Listener/foo/bar/*"  
"xdstp://some-authority/envoy.config.listener.v3.Listener/foo/baz/*"]

htuch:
Sorry, this has idled a bit, I will update today.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: I'm going to drop both DCPs _and_ client feature capabilities from this xRFC and leave them for your dynamic context params extensions.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1067 @@
...
+    transformation](https://developers.google.com/protocol-buffers/docs/proto3#json).
+  * Both `xds.node.metadata.X` and
+    `xds.node.user_agent_build_version.metadata.X` permit directly referencing
+    a top-level metadata field `X` in a context parameter key. This does not
+    apply to nested fields in the metadata, e.g. `xds.node.metadata.bar`.
+  * `xds.node.user_agent_build_version.version` yields a string value composed
+    of `major.minor.patch` values, e.g. `"1.2.3"`.
+
+* Context parameters from the URL, in the above example
+  `xds.bar=baz`. These must not overlap with any other context parameter name,
```
htuch, line 295: @markdroth should we put this in some namespace, e.g. `xds.params.` to future proof other parts of the `xds.` hierarchy in case we want them?

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,1067 @@
...
+  have `{xds.resource.vip: 96.54.3.1}`. These attributes are generated by code,
+  e.g. an on-demand LDS implementation will first detect the VIP, halt the
+  request, ask for the `xds.resource.vip` from the control plane, and then
+  resume the request after discovery is completed. All these attributes will be
+  `xdstp.resource.` prefixed.
+
+An example computed URN following the above example is
+`xdstp://some-authority/some.type/foo?xds.node.metadata.foo=bar&xds.shard_id=1234&xds.bar=baz&xds.client_feature.lb.least_loaded=false&xds.resource.vip=96.54.3.1`.
+
+This proposal reserves the `@` prefix for context parameter values for future
```
htuch, line 308: @markdroth should we still keep this one?

proposals/TP1-xds-transport-next.md
htuch: So, if we stick with what is here and either spell out the bootstrap `ConfigSource` style or leave to another PR this is fine?

I'm happy to dump the gRPC opinionated bootstrap schema here and adopt, as I think we want this in Envoy as well. Do you have examples?

htuch:
@markdroth @fuqianggao @mattklein123 @norafang52 any further feedback or can we merge? I'd like to get the baseline RFC landed for us to iterate later.

mattklein123:
LGTM to ship and iterate.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,864 @@
+TP1: xdstp:// structured resource naming, caching and federation support.
+----
+* Author(s): Harvey Tuch, Louis Ryan, Mark Roth, Costin Manolache, Matt Klein
+* Approver:
+* Implemented in: <xDS client, ...>
+* Last updated: 2021-04-12
```
markdroth, line 6: Please update the date before merging. :)

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,864 @@
...
+semantics by examining the prefix of a resource string.
+
+Clients are expected to have knowledge ahead of time (via mechanisms not part of
+this proposal, e.g. support documentation) on whether the servers specified in
+the bootstrap supports the new URI conventions.
+
+### Server redirects
+
+The server may issue redirects by including an `xds::core::v3::ResourceLocator`
+in a `Resource` wrapped in the `Any` message. Redirects are only supported for
```
markdroth, line 530: Does it still make sense to use `xds.core.v3.ResourceLocator` on the wire for redirects, since we're no longer using that proto message for anything else on the wire?  I'm wondering if we should instead just have something simple like this:

```
message Redirect {
  string resource_locator = 1;
}
```

Then a redirect would be triggered by having the `Resource` include a message of type `xds.core.v3.Redirect`.

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,864 @@
...
+- version: 13
+  name: xdstp://some-authority/envoy.config.listeners.v3.Listener/b-listeners/baz
+  resource:
+    "@type": type.googleapis.com/envoy.config.listeners.v3.Listener
+    … baz's Listener payload … 
+```
+
+## Rationale
+
+[A discussion of alternate approaches and the trade offs, advantages, and disadvantages of the specified approach.]
```
markdroth, line 855: We should probably either fill in these last few sections or else replace the contents with "N/A". :)

markdroth (commented):
This looks really good overall!  I have just a few comments.

Please let me know if you have any questions.  Thanks!

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,864 @@
...
+
+Since each resource returned is subject to independent update via delta xDS and
+there is no explicit collection directory to update, glob collections are highly
+scalable. For example, consider a collection of 10k resources and the addition
+of a single resource. Glob collections will send the single additional resource
+(due to the operation of delta xDS). List collections requires that the
+collection directory (now containing 10001 resource references) to be sent from
+server to client, as well as the additional new resource.
+
+A server may reply with a sentinel `Resource` with no value populated and
```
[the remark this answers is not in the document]
adisuissa: I see how this can work for many examples.

It might be my misinterpretation of `removed_resources`, but I might have a use-case that this doesn't cover:
1. Client's bootstrap is set to fetch a glob collection of clusters:
```
dynamic_resources:
  cds_resources_locator:
    name: xdstp://some-authority/envoy.config.clusters.v3.Cluster/my-clusters/*
  cds_config:
    authorities:
    - name: some-authority
```
2. At the moment the (due to some transient error) the management server has no clusters configured, and so it responds with `removed_resources:["xdstp://some-authority/envoy.config.clusters.v3.Cluster/my-clusters/*"]`.
3. The client receives the response and is no longer registered to that glob list.

I think that the sentinel allows to keep the registration alive.
WDYT?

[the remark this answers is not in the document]
markdroth: Adi, the server telling the client that the resource does not exist does not cause the client to unsubscribe from the resource.  Only the client controls what resources it is subscribed to; the server cannot alter that list.  The client will still be subscribed, and if the resource is created later, the server will send an update containing the resource.

You can think of "resource does not exist" as just a special sentinel value of a given resource.  While the client is subscribed to a resource, the server sends updates whenever the resource changes from one value to another, and "does not exist" is just another possible value for the resource.  Whenever the resource's value changes from some other value to "does not exist" or from "does not exist" to some other value, the server will send an update, just like it would on any other value change.

proposals/TP1-xds-transport-next.md
htuch: OK, I'm adding just a simple `repeated` to give an idea of what is possible and will quality the dynamic `ConfigSource` is not necessary if a client has another resolution mechanism.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: Added a deferral sentence.

proposals/TP1-xds-transport-next.md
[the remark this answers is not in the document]
htuch: Updated to this scheme, I don't see a big deal either way.

proposals/TP1-xds-transport-next.md
```diff
@@ -424,9 +423,9 @@ of a single resource. Glob collections will send the single additional resource
 collection directory (now containing 10001 resource references) to be sent from
 server to client, as well as the additional new resource.
 
-A server may reply with a sentinel `Resource` with no value populated and
-resource name matching the glob collection in returned resources to indicate
-that the collection resource set is currently empty.
+If no resources are present in the glob collection, the server should reply with a
+`DeltaDiscoveryResponse` in which the glob collection URL is specififed in
```
markdroth, line 427: s/specififed/specified/

markdroth (approved):
Just one small typo to fix, otherwise this looks fantastic!

proposals/TP1-xds-transport-next.md
```diff
@@ -0,0 +1,873 @@
...
+expected to be slow on the uptake and continued SotW support is provided for
+this proposal.
+
+### Server redirects
+
+The server may issue redirects by including the following message
+in a `Resource` wrapped in the `Any` message:
+
+```
+message Redirect {
```
markdroth, line 539: One more quick change here.  Let's change this to:

```
message ResourceLocator {
  string name = 1;
}
```

That way, we can use the same message for both redirects and for subscriptions/unsubscriptions, since we'll be making a similar change for subscriptions/unsubscriptions as part of the dynamic parameter design.

adisuissa (approved):

markdroth (approved):
Looks great!  I think it's ready to merge.

PR #10, opened 2021-08-04
xRFC TP2: Dynamically Generated Cacheable xDS Resources

markdroth:
CC @fuqianggao @adisuissa

### DIFF
--- a/proposals/TP2-dynamically-generated-cacheable-xds-resources.md
+++ b/proposals/TP2-dynamically-generated-cacheable-xds-resources.md
@@ -1,0 +1,1064 @@
+TP2: Dynamically Generated Cacheable xDS Resources
+----
+* Author(s): markdroth, htuch
+* Approver: htuch
+* Implemented in: <xDS client, ...>
+* Last updated: 2021-08-03
+
+## Abstract
+
+This xRFC proposes a new mechanism to allow xDS servers to
+dynamically generate the contents of xDS resources for individual
+clients while at the same time preserving cacheability.  Unlike the
+context parameter mechanism that is part of the new xDS naming scheme (see
+[xRFC TP1](https://github.com/cncf/xds/pull/6)), the mechanism described in
+this proposal is visible only to the transport protocol layer, not to the
+data model layer.  This means that if a resource has a parameter that
+affects its contents, that parameter is not part of the resource's name,
+which means that any other resources that refer to the resource do not
+need to encode the parameter.  Therefore, use of these parameters is
+not viral, thus making the mechanism much easier to use.
+
+## Background
+
+There are many use-cases where a control plane may need to dynamically
+generate the contents of xDS resources to tailor the resources for
+individual clients.  Here are some examples:
+
+- **xDS minor/patch version negotiation.** In this case, each client
+  supports a given minor and patch version, and the server may choose to
+  use newer API features when talking to clients that support newer versions
+  of the API.  (See https://github.com/envoyproxy/envoy/issues/8416 for
+  details.)
+- **Sharding Cluster resources for scalability.**  In this use-case, there
+  are a really large number of clusters, too many for any one client to
+  handle.  The clusters are divided into shards, and each client is given
+  a dynamically changing assignment of which shards to load.  To support
+  this, there needs to be a different variant of the `ClusterCollection`
+  resource for each combination of shards that may be assigned to a given
+  client.  The shard assignments are generally determined dynamically on
+  the client but may change at any time.
+- **Sharding endpoints for scalability.**  This is similar to the previous
+  case, except that there is a single cluster with a large number of
+  endpoints.  The goal is that the xDS server will send different subsets
+  of endpoints to different clients, thus avoiding unwanted connections
+  when there are large numbers of both servers and clients.  (At Google,
+  this is referred to as "subsetting", but it's a different feature than
+  the one that Envoy uses that term for.)  In this case, it is desirable
+  for the xDS server to determine the subset of endpoints to assign to
+  each client.
+- **Selecting which cluster to send a client to based on an ACL.**  In this
+  use-case, there are two different network paths that can be used to
+  access the endpoints: one goes directly to the endpoints, with
+  client-side load balancing, and the other goes via a reverse proxy.
+  The path that goes directly to the endpoints is faster but is
+  access-restricted.  The xDS server needs to check an ACL to determine
+  whether a given client is authorized to directly access the endpoints.
+  If the client is authorized, it will be sent a `RouteConfiguration`
+  pointing to the cluster for those endpoints; otherwise, it will be sent
+  a different variant of the `RouteConfiguration` that points to a cluster
+  containing the reverse proxy endpoint.
+- **Dynamic route selection.**  Every client sends a set of dynamic
+  selection parameters (today, conveyed as node metadata).  The server
+  has a list of routes to configure, but individual routes in the list
+  may be included or excluded based on the client's dynamic selection
+  parameters.  Thus, the server needs to generate a slightly different
+  version of the `RouteConfiguration` for clients based on the parameters
+  they send.  (See
+  https://cloud.google.com/traffic-director/docs/configure-advanced-traffic-management#config-filtering-metadata
+  for an example.)
+
+The new xDS naming scheme described in [xRFC
+TP1](https://github.com/cncf/xds/pull/6) provides a mechanism called
+context parameters, which is intended to move all parameters that affect
+resource contents into the resource name, thus adding cacheability to the
+xDS ecosystem.  However, this approach means that these parameters
+become part of the resource graph on an individual client, which causes
+a number of problems:
+- Dynamic context parameters are viral, spreading from a given resource
+  to all earlier resources in the resource graph.  For example, if
+  multiple variants of an EDS resource are needed, there need to be two
+  different instances of the resource with different names,
+  distinguished by a context parameter.  But because the contents of the
+  CDS resource include the name of the corresponding EDS resource name,
+  that means that we also need two different versions of the CDS
+  resource, also distinguished by the same context parameter.  And then
+  we need two different versions of the RDS resource, since that needs
+  to refer to the CDS resource.  And then two different versions of the
+  LDS resource, which refers to the RDS resource.  This causes a
+  combinatorial explosion in the number of resources needed, and it adds
+  complexity to xDS servers, which need to construct the right variants
+  of every resource and make sure that they refer to each other using
+  the right names.
+- In the new xDS naming scheme, context parameters are exact-match-only.
+  This means that if a control plane wants to provide the same resource
+  both with and without a given parameter, it needs to publish two
+  versions of the resource, each with a different name, even though the
+  contents are the same, which can also cause unnecessarily poor cache
+  performance.  For example, in the "dynamic route selection" use-case,
+  let's say that every client uses two different dynamic selection
+  parameters, `env` (which can have one of the values `prod`, `canary`, or
+  `test`) and `version` (which can have one of the values `v1`, `v2`, or
+  `v3`).  Now let's say that there is a `RouteConfiguration` with one route
+  that should be selected via the parameter `env=prod` and another route that
+  should be selected via the parameter `version=v1`. This means that there
+  are only four variants of the `RouteConfiguration` resource (`{env!=prod,
+  version!=v1}`, `{env=prod, version!=v1}`, `{env!=prod, version=v1}`, and
+  `{env=prod, version=v1}`).  However, the exact-match semantics means
+  that there will have to be nine different versions of this resource,
+  one for each combination of values of the two parameters.
+
+### Related Proposals:
+* [xRFC TP1: new xDS naming scheme](https://github.com/cncf/xds/pull/6)
+
+## Proposal
+
+This document proposes an alternative approach.  We start with the
+observation that resource names are used in two places:
+
+- The **transport protocol** layer, which needs to identify the right
+  resource contents to send for a given resource name, often obtaining
+  those resource contents from a cache.
+- The **resource graph** used on an individual client, where there are a
+  set of data model resources that refer to each other by name.  For
+  example, a `RouteConfiguration` refers to individual `Cluster` resources
+  by name.
+
+The use-cases for dynamic resource selection share one important property
+that we can take advantage of.  When multiple variants of a given resource
+exist, any given client will only ever use one of those variants at a
+given time.  That means that the parameters that affect which variant
+of the resource is used are required by the transport protocol, but
+they are not required by the client's data model.  (For example, in the
+"sharding endpoints for scalability" use-case, different clients may see
+different variants of the EDS resource, but once a given client has the
+right variant, it will be unique on that client, which means that the
+CDS resource does not need to refer to different EDS resource names on
+different client.)
+
+It should be noted that caching xDS proxies, unlike "leaf" clients, will
+need to track multiple variants of each resource, since a given caching
+proxy may be serving clients that need different variants of a given
+resource.  However, since caching xDS proxies deal with resources only
+at the transport protocol layer, the resource graph layer is
+essentially irrelevant in that case.
+
+### Dynamic Parameters
+
+With the above property in mind, this document proposes the following
+data structures:
+- **Dynamic parameters**, which are a set of key/value pairs that are part
+  of the cache key for an xDS resource (in addition to the resource name
+  itself).  This provides a mechanism to represent multiple variants of a
+  given resource in a cacheable way.  These parameters are used to identify
+  the specified resource in the transport protocol, but they are not part of
+  the resource name and therefore do not appear as part of the resource graph.
+- **Dynamic parameter constraints**, which are a set of criteria that
+  can be used to determine whether a set of dynamic parameters matches
+  the constraints.  When a client subscribes to a resource, it may
+  specify a set of dynamic parameter constraints, which will be used to
+  select which variant of the resource will be returned by the server.
+  In response to a given subscription request from the client containing
+  a set of dynamic parameter constraints, the server will send a
+  resource whose dynamic parameters match the dynamic parameter
+  constraints in the request.  The client will use the dynamic
+  parameters on the resource to determine which of its subscriptions the
+  resource is associated with.
+
+Dynamic parameters, unlike context parameters, will not be
+exact-match-only.  Dynamic parameter constraints will be able to represent
+various types of flexible matching, such as range-based matching (which
+will be used for the "xDS minor/patch version negotiation" use-case).
+This flexible matching semantic means that there are some cases where
+ambiguity can occur; we define a set of best practices below to prevent
+these cases from occurring in practice.
+
+#### Matching Ambiguity
+
+Flexible matching means that there may be ambiguities when determining
+which resources match which subscriptions.  This section defines the matching
+behavior and a set of best practices for deployments to follow to avoid this
+kind of ambiguity.
+
+To illustrate where this comes up in practice, it is useful to consider
+what happens in transition scenarios, where a deployment initially
+groups its clients on a single key but then wants to add a second key.
+The second key needs to be added both in the constraints on the server
+side and in the clients' configurations, but those two changes cannot
+occur atomically.
+
+For example, let's say that the clients are currently categorized by the
+parameter `env`, whose value is either `prod` or `test`.  The resource
+variants on the server will therefore have the following sets of dymamic
+parameters:
+- `{env=prod}`
+- `{env=test}`
+
+Clients will send one of the following two sets of dynamic parameter
+constraints, depending on whether they are `prod` or `test` clients:
+
+```textproto
+// For {env=prod}
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"prod"}]
+  }}
+]}
+
+// For {env=test}
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"test"}]
+  }}
+]}
+```
+
+Now the deployment wants to add an additional key called `version`,
+whose value will be either `v1` or `v2`, so that it can further subdivide
+its clients' configs.
+
+If the new key is added on the clients first, then the clients will
+start subscribing with dynamic parameters constraints like the following:
+
+```textproto
+// For {env=prod, version=v1}
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"prod"}]
+  }},
+  {key:"version" value:{
+    constraints:[{value:"v1"}]
+  }}
+]}
+```
+
+The server or cache has to match that set of constraints against the
+existing sets of dynamic parameters, which do not specify the `version`
+key at all.
+
+Conversely, if the new key is added on the server side first, then the
+server will have resource variants with parameters like this:
+- `{env=prod, version=v1}`
+- `{env=prod, version=v2}`
+- `{env=test, version=v1}`
+- `{env=test, version=v2}`
+
+But at this point, the clients are continuing to subscribe without
+constraints on this new key.  So the server or cache needs to figure out
+(e.g.) which of the first two sets of constraints to use for constraints
+that require `env` to be `prod` but do not specify `version`.
+
+We address this transition scenario by allowing the set of constraints
+for a given key to match any resource variant that does not specify that
+key at all.  This allows constraints for new keys to be added on clients
+before the corresponding keys are added on the resources on the server, but
+it does introduce some additional ambiguity into the matching.  For example,
+let's say that the server has the following two variants of a resource:
+- `{env=prod}`
+- `{env=prod, version=v1}`
+
+Now consider what happens if a client subscribes with the following
+constraints:
+
+```textproto
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"prod"}]
+  }},
+  {key:"version" value:{
+    constraints:[{value:"v1"}]
+  }}
+]}
+```
+
+These constraints can match either of the above variants of the resource.
+This situation can be avoided by establishing a best practice that all
+variants of a given resource must have the same set of keys.
+
+There is still a possible ambiguity that can occur if a server adds
+multiple variants of a new key that clients are not yet sending.
+For example, let's say that the server has the following two variants
+of a resource:
+- `{env=prod, version=v1}`
+- `{env=prod, version=v2}`
+
+Consider what happens if a client subscribes with the following constraints:
+
+```textproto
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"prod"}]
+  }}
+]}
+```
+
+These constraints can match either variant of the above resource.
+This can be avoided by establishing a best practice of not adding multiple
+variants of a new parameter until clients are sending the new parameter.
+However, if this does happen, the cache implementation is free to pick
+one of the variants at random.
+
+So, the expected order of changes for this kind of transition would be:
+1. Change clients to start sending a constraint for `version=v1`.
+2. Add the dynamic parameter `version=v1` to all existing resources.
+3. Create new variants of each resource with `version=v2`.
+4. Change the desired set of clients to send a constraint for
+   `version=v2` instead of `version=v1`.
+
+##### Alternatives Considered
+
+We could avoid much of the matching ambiguity described above by saying that
+a set of constraints must specify all keys present on the resource in order
+to match.  However, this would mean that if the client starts subscribing
+with a constraint for the new key before the corresponding key is added on
+the resources on the server, then it will fail to match the existing resources.
+In other words, the process would be:
+
+1. Add a variant of all resources on the server side with `version=v1`
+   (in addition to all existing dynamic parameters).
+2. Change clients to start sending constraints with the new key.
+3. When all clients are updated, remove the resource variants that do
+   *not* have the new key.
+
+This will effectively require adding new keys on the server side first,
+which seems like a large burden on users.  It also seems fairly tricky
+for most users to get the exactly correct set of dynamic parameters on
+each resource variant, and if they fail to do it right, they will break
+their existing configuration.
+
+We also considered having the client add the new constraint but mark it
+as optional using an `is_optional` field.  That way, it would match
+resources both before and after the new key is added on the server.
+However, the `is_optional` field would introduce another type of ambiguity
+in matching.  Specifically, let's say that the server has the following
+two variants of the resource:
+- `{env=prod}`
+- `{env=prod, version=v1}`
+
+Now a client subscribes with the following set of constraints:
+
+```textproto
+{key_constraints:[
+  {key:"env" value:{
+    constraints:[{value:"prod"}]
+  }},
+  {key:"version" value:{
+    constraints:[{value:"v1"}]
+    is_optional: true
+  }}
+]}
+```
+
+These constraints can match either of the above variants of the resource.
+For authoritative servers, this could be addressed by establishing
+a best practice of not having two variants of a resource that differ
+only by keys that the client will send as optional.  However, this
+requires coordination between client and server, and requires
+machinery on the client to determine when to set the `is_optional` bit.
+
+Ultimately, although this approach is more semantically precise, it is
+also considered too rigid and difficult for users to work with.
+
+#### Matching Behavior and Best Practices
+
+We advise deployments to avoid ambiguity through the following best practices:
+- Whenever there are multiple variants of a resource, all variants must
+  list the same set of keys.  This allows the server to ignore constraints
+  on keys sent by the client that do not affect the choice of variant
+  without causing ambiguity in cache misses.
+- Servers should not create multiple variants of a parameter that is not yet
+  being sent by clients.  If they do, clients that do not send that parameter
+  will get one of the variants at random.
+- There must be a variant of the resource for every value of a key that is
+  going to be present.  For example, if clients will send constraints on the
+  `env` key requiring the value to be one of `prod`, `test`, or `qa`, then
+  you must have each of those three variants of the resource.
+  - Note that servers that can make use of the mechanism described under
+    [Server-Specified Constraints](#server-specified-constraints) below
+    may be able to optimize this in some cases.  See the "Dynamic Route
+    Selection" example below for details.
+- For cases where a constraint may match multiple values (e.g., a
+  range constraint), the largest possible matching value is preferred.
+  This means that caches (both on clients and on caching xDS proxies)
+  must attempt to fetch a larger value even if they already have a smaller
+  matching value already present in the cache.  For example, let's say
+  that a cache contains a variant of a resource with the parameter
+  `{shard=3}` and a client subscribes with the following constraints:
+  ```textproto
+  {key_constraints:[
+    {key:"shard" value:{
+      constraints:[
+        {integer_range_list:[
+          {range:{min_value:0 max_value:5}}
+        ]}
+      ]
+    }}
+  ]}
+  ```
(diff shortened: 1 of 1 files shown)


howardjohn:
@hzxuzhonghu @ramaraochavali  should take a look as we have been looking into caching XDS in Istio. I haven't had a chance to take a look yet but planning to soon.

hzxuzhonghu:
What's tricky for istio is: we donot know which are the dynamic parameters for xds clients in advance, because users may have policies with workload selector.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,1066 @@
...
+    constraints:[{value:"prod"}]
+  }}
+]}
+```
+
+These constraints can match either variant of the above resource.
+This can be avoided by establishing a best practice of not adding multiple
+variants of a new parameter until clients are sending the new parameter.
+However, if this does happen, the cache implementation is free to pick
+one of the variants at random.
```
htuch, line 298: Arguably we should suggest it would be preferable to make this deterministic for any given client and known server state, i.e. some stability property. It's still not great though, since a client reconnect to a different server will result in something different.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,1066 @@
...
+In other words, the process would be:
+
+1. Add a variant of all resources on the server side with `version=v1`
+   (in addition to all existing dynamic parameters).
+2. Change clients to start sending constraints with the new key.
+3. When all clients are updated, remove the resource variants that do
+   *not* have the new key.
+
+This will effectively require adding new keys on the server side first,
+which seems like a large burden on users.  It also seems fairly tricky
```
htuch, line 323: Why is this so bad? We basically need to roll out somewhere first.

htuch (commented):
I've read up to the API protos, looks good, but would like to discuss a little bit on the tradeoffs around matching semantics for missing labels.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,1066 @@
...
+These constraints can match either variant of the above resource.
+This can be avoided by establishing a best practice of not adding multiple
+variants of a new parameter until clients are sending the new parameter.
+However, if this does happen, the cache implementation is free to pick
+one of the variants at random.
+
+So, the expected order of changes for this kind of transition would be:
+1. Change clients to start sending a constraint for `version=v1`.
+2. Add the dynamic parameter `version=v1` to all existing resources.
+3. Create new variants of each resource with `version=v2`.
```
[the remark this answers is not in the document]
markdroth: I think so, although @fuqianggao can chime in as well.

I would expect the normal use-case here to be that the user configures their clients to start sending a constraint for the new key before they start differentiating their resources based on that key.  Doing the reverse doesn't really make sense, because there isn't yet anything to differentiate.

To say this another way, I think it could be quite common for users to divide their clients into categories based on some key, even if they initially don't configure those sets of clients differently from each other.  But the fact that they have that in place means that they can easily change the config of a subset of their clients simply by changing the configuration reesources in the control plane.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: If we think users can deal with this, I'd actually prefer this approach, because it would basically eliminate the ambiguity problem.  But from what @fuqianggao has said, I believe that one of the common use-cases is the one I mentioned above, where users divide their clients into categories based on some key, even if they initially don't configure those sets of clients differently from each other, so that they can easily change the config of a subset of their clients later simply by changing the configuration reesources in the control plane.  I'm not sure how to support that use-case with requiring all keys to exist.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: Yeah, I thought about that, but I don't think it will do much good.  This same matching needs to be done not just on the authoritative server but also in caching proxies and in the cache on leaf clients.  And there is no guarantee that the caches will have the same set of resources as the authoritative server, so they generally won't be able to make the same choice anyway.

I think the only way to get deterministic behavior is to require users to deploy the change starting at the clients.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,1066 @@
...
+
+Dynamic parameters, unlike context parameters, will not be
+exact-match-only.  Dynamic parameter constraints will be able to represent
+various types of flexible matching, such as range-based matching (which
+will be used for the "xDS minor/patch version negotiation" use-case).
+This flexible matching semantic means that there are some cases where
+ambiguity can occur; we define a set of best practices below to prevent
+these cases from occurring in practice.
+
+#### Matching Ambiguity
```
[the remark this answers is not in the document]
markdroth: I've moved the baseline example from below up, so hopefully this reads a bit better now.

markdroth:
I've removed the DPDS functionality from this xRFC, since I don't think it fully solves the problem of server-specified constraints: the DPDS resource itself would be uncacheable, which means that the client needs to talk to the authoritiative server anyway, but the whole point of dynamic parameters is to make resources cacheable.  Therefore, DPDS seems no better than simply marking the resources that we wanted to dynamically generate as being non-cacheable.

I do think this issue will probably need more attention later, but I think we can decouple that from the basic dynamic parameter design here.

adisuissa (commented):
Thanks Mark, this is a great!
Left a few comments.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,1066 @@
...
+
+## Abstract
+
+This xRFC proposes a new mechanism to allow xDS servers to
+dynamically generate the contents of xDS resources for individual
+clients while at the same time preserving cacheability.  Unlike the
+context parameter mechanism that is part of the new xDS naming scheme (see
+[xRFC TP1](TP1-xds-transport-next.md)), the mechanism described in
+this proposal is visible only to the transport protocol layer, not to the
+data model layer.  This means that if a resource has a parameter that
```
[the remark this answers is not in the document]
htuch: I think I will just have to practice acceptance and equanimity in this case then :)

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
htuch: Makes sense, I don't disagree with the theory here, I'm thinking maybe something empirical to back this from TD as an example would make me more confident.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,722 @@
...
+
+Today, the equivalent of dynamic parameter constraints is node metadata,
+which can be used by servers to determine the set of resources to send
+for LDS and CDS wildcard subscriptions or to determine the contents of
+other resources (e.g., to select individual routes to be included in an
+RDS resource).  For transition purposes, this mechanism can continue
+to be supported by the client performing direct translation of node
+metadata to exact-match constraints.  For example, if the node metadata
+contains the entry `env=prod`, this would be translated to a constraint
+`{key_constraints:[{key:"env" value:{constraints:[{value:"prod"}]}}]}`.
```
htuch, line 518: How do we know that `env` is from node metadata? I think this might matter in some cases, e.g. when the same key is used in node metadata and real DCP.

htuch (commented):
At high-level looks great. I can do a nit pick pass again when threads are wrapped.

ejona86 (commented):
I've not read every word of the proposal, so I apologize if something was discussed and I missed it.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+  repeated ConstraintList constraints = 1;
+}
+```
+
+#### Matching Behavior
+
+Note that both xDS servers and clients need to evaluate matching between
+a set of dynamic parameters and a set of constraints.  The server does
+this when deciding which variant of a given resource to return for a
+given subscription request.  When the client receives the resource from
```
ejona86, line 185: The client only needs to do matching behavior if it has multiple watches for the same resource with varying dynamic parameters. It was already mentioned that clients will generally only have one set of dynamic parameters for a resource (which I assume to be immutable), so it seems the common case is that clients don't need to do the constraint processing.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+client capability.
+
+Because of this, reviewers of this design are encouraged to carefully
+scrutinize the proposed matching semantics to ensure that they meet our
+expected needs.
+
+### Using Context Parameters
+
+We considered extending the context parameter mechanism from [xRFC
+TP1](TP1-xds-transport-next.md) to support flexible matching semantics,
```
ejona86, line 652: So if this proposal is approved, TP1 will really be serving federation and this TP2 will be serving caching?

How does this interact with federation? I don't see a mention to the "scope" that dynamic parameters are configured on client-side. That's fair from the transport-protocol perspective, but makes it hard to understand "does it solve the appropriate use cases."

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+### Migrating From Node Metadata
+
+Today, the equivalent of dynamic parameter constraints is node metadata,
+which can be used by servers to determine the set of resources to send
+for LDS and CDS wildcard subscriptions or to determine the contents of
+other resources (e.g., to select individual routes to be included in an
+RDS resource).  For transition purposes, this mechanism can continue
+to be supported by the client performing direct translation of node
+metadata to exact-match constraints.  For example, if the node metadata
+contains the entry `env=prod`, this would be translated to a constraint
```
ejona86, line 520: This seems broken. I expected `env=prod` to be translated to dynamic parameters `{env=prod}` on client-side if we're talking about doing translation. But this is talking about _constraints_ which means it is server-side. It seems clients/proxies wouldn't be able to match these constraints in response objects, because their dynamic properties _don't_ include `env`.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+]}
+```
+
+Now consider what happens if a client subscribes with dynamic parameters
+`{env=test}`.  Those dynamic parameters can match either of the above
+variants of the resource.
+
+This situation will be avoided by requiring that **all variants of a given
+resource must specify non-overlapping constraints for the same set of keys**.
+Control planes must not accept a set of resources that violates this
```
ejona86, line 436: This seems suspiciously similar to the 3-SAT NP-complete problem. A control plane could side-step that issue by giving a reduced configuration format and having limits, but is that what we had in mind? If so, we should provide some warning here "beware the computation complexity".

I honestly don't see the point of this requirement. The client already is supposed to handle ambiguity by selecting a resource arbitrarily. I could understand this requirement more if the client could assume there will never be ambiguity. Or if it is a "best practice" instead of "requirement."

markdroth (commented):
Thanks for the detailed review, Eric!

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+      // If set to true, the match is inverted -- i.e., the key must NOT
+      // match the specified value.
+      bool invert = 4;
+    }
+
+    enum MatchType {
+      // Default value.
+      MATCH_TYPE_UNSPECIFIED = 0;
+      // Logical AND of constraints.
+      MATCH_TYPE_AND = 1;
```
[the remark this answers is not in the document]
markdroth: Sorry, forgot to respond to this one.

This is a good point.  I'd like some input from @fuqianggao on this, since this was based on a similar internal proto that uses this same format.  If it's really not useful, I'm happy to get rid of it, but I'd like to first verify that this wasn't added for some particular reason internally.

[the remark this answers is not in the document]
fuqianggao: I think it is for better UX.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,759 @@
...
+  example, a `RouteConfiguration` refers to individual `Cluster` resources
+  by name.
+
+The use-cases for dynamic resource selection share one important property
+that we can take advantage of.  When multiple variants of a given resource
+exist, any given client will only ever use one of those variants at a
+given time.  That means that the parameters that affect which variant
+of the resource is used are required by the transport protocol, but
+they are not required by the client's data model.  (For example, in the
+"sharding endpoints for scalability" use-case, different clients may see
```
fuqianggao, line 96: I thought this use case is out of context for this design?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,722 @@
...
+will be used for the "xDS minor/patch version negotiation" use-case).
+This flexible matching semantic means that there are some cases where
+ambiguity can occur; we define a set of best practices below to prevent
+these cases from occurring in practice.
+
+For example, let's say that the clients are currently categorized by the
+parameter `env`, whose value is either `prod` or `test`.  The resource
+variants on the server will therefore have the following sets of dymamic
+parameters:
+- `{env=prod}`
```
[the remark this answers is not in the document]
markdroth: In this particular example, those are the only two variants, there are no others.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: Good catch, this is a lingering comment from an earlier draft.  Removed.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
markdroth: Can you say more about why you think it's better?  It's not obvious to me what the benefit is.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+Ultimately, although this approach is more semantically precise, it is
+also considered too rigid and difficult for users to work with.
+
+## Implementation
+
+TBD (Will probably be implemented in gRPC before Envoy)
+
+## Open issues (if applicable)
+
+N/A
```
htuch, line 754: One thing I'm missing in this xRFC is discussion on caching data structures. Since they are contraint based, this is not in general a simple map. Can we provide some recommended algorithms or reference pseudo-code?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+message DynamicParameterConstraints {
+  // A list of constraints that may be combined with AND or OR semantics.
+  message ConstraintList {
+    // A constraint for a given key.
+    message Constraint {
+      message Exists {}
+      // The key to match against.
+      string key = 1;
+      // How to match.
+      oneof constraint_type {
```
htuch, line 150: If we end up growing this in the future, we will need to teach client, server and intermediate layers about the new proto. Do we need any kind of version negotiation or reporting etc? How would this work in practice?

htuch (changes requested):
Overall proposal looks good, a bunch of tiny and bigger comments. Thanks for putting this together @markdroth!
/wait

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+
+Now let's say that the server has two variants of a given resource, and
+the variants have the following dynamic parameter constraints:
+
+```textproto
+// For {env=prod}
+{constraints:[
+  {
+    constraints:[{key:"env" value:"prod"}]
+    match_type: MATCH_TYPE_AND
```
mattklein123, line 228: nit: it's kind of confusing that we specify AND here with a single constraint. Maybe remove and/or say this doesn't need setting with a single value? Up to you.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+
+This will work fine for older clients that are not yet sending the
+`version` key, because their dynamic parameters will not match the new
+variant's constraints.  However, newer clients that are sending dynamic
+parameters `{env=prod, version=v1}` will run into ambiguity: those
+parameters can match either of the above variants of the resource.
+
+This situation will be avoided via a best practice that all authoritative
+xDS servers should have **all variants of a given resource specify
+constraints for the same set of keys**.
```
mattklein123, line 402: Thinking out loud, couldn't this case also be handled by recommending that the server return the most specific match (the most keys that match)? This would not handle overlapping cases, but would handle this case well.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+```
+
+Now consider what happens if a client subscribes with dynamic parameters
+`{env=test}`.  Those dynamic parameters can match either of the above
+variants of the resource.
+
+This situation will be avoided via a best practice that all authoritative
+xDS servers should have **all variants of a given resource specify
+non-overlapping constraints for the same set of keys**.  Control planes
+must not accept a set of resources that violates this requirement.
```
mattklein123, line 479: Thinking out loud. Isn't this really up to the deployment? Meaning, this can be the recommendation, but wouldn't it also be fine to have some kind of internal ordering/priority that would kick in if there is an overlap?

mattklein123 (changes requested):
Sorry for the really, really, really late review. Overall this looks good. I left some comments. Thank you!

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+xDS proxies; a proxy may have multiple clients that request different
+variants of the same resource.  Because we do not want to get into a
+situation where xDS servers typically do not populate dynamic parameter
+constraints in their responses and then need changes to work with
+caching proxies, this design requires that leaf clients validate that the
+constraints in the response match the requested dynamic parameters.
+This ensures that the wire protocol used by leaf clients and caching xDS
+proxies remains the same.
+
+(In effect, the resource cache in an xDS client is basically the same
```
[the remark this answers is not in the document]
markdroth: I've restructured this a bit and split it into two sections.  The first section provides background on xDS client and server architecture, and the second section describes how dynamic parameter matching fits into that architecture.

I'm now wondering if maybe I should copy some of that first section into https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol...  Thoughts?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+of a given resource, implementations may be tempted to not bother with
+this matching on the client side.  However, that is not true for caching
+xDS proxies; a proxy may have multiple clients that request different
+variants of the same resource.  Because we do not want to get into a
+situation where xDS servers typically do not populate dynamic parameter
+constraints in their responses and then need changes to work with
+caching proxies, this design requires that leaf clients validate that the
+constraints in the response match the requested dynamic parameters.
+This ensures that the wire protocol used by leaf clients and caching xDS
+proxies remains the same.
```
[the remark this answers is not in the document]
markdroth: I've changed this to require servers to send the dynamic parameter constraints, and I've explicitly pointed out that leaf clients may or may not verify the constraints, but that servers that don't send them will get into trouble later when they want to start supporting caching proxies.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+      MATCH_TYPE_AND = 1;
+      // Logical OR of constraints.
+      MATCH_TYPE_OR = 2;
+    }
+
+    // A list of key/value constraints.
+    repeated Constraint constraints = 1;
+
+    // How to match the constraints.
+    MatchType match_type = 2;
```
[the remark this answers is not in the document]
markdroth: I think the trade-off here is that the more complex we make this, the more flexible it is, but the harder it is for caches (in both clients and servers) to optimize the matching implementation.  See the section I added below on implementation considerations for some thoughts on this.

That having been said, given that this will be hard to extend later, I do agree that if we think we'll need something, we should make sure to add it now.  We just need to figure out where the right balance is, since we're making a trade-off here.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+```
+
+#### Where Matching Is Performed
+
+Both xDS servers and clients need to evaluate matching between a set
+of dynamic parameters and a set of constraints.  The server does this
+when deciding which variant of a given resource to return for a given
+subscription request.  When the client receives the resource from
+the server, it needs to do the same matching to determine which of its
+subscriptions that resource is associated with.  Therefore, the matching
```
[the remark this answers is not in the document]
markdroth: I've restructured this section, splitting it into two.  Hopefully it's clearer now.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+behavior becomes an inherent part of the xDS transport protocol.
+
+Note that because leaf clients should only ever receive a single variant
+of a given resource, implementations may be tempted to not bother with
+this matching on the client side.  However, that is not true for caching
+xDS proxies; a proxy may have multiple clients that request different
+variants of the same resource.  Because we do not want to get into a
```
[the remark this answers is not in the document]
markdroth: Restructured to try to make this clearer.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: I think this is related to the question above about whether `MATCH_TYPE_AND` is actually useful at all.  If it's not, we can remove it, but I'd like @fuqianggao to weigh in on this first, since it was based on an internal proto that he showed me.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,754 @@
...
+      {key:"env" value:"prod"},
+      {key:"version" value:"v2"}
+    ]
+    match_type: MATCH_TYPE_AND
+  }
+]}
+```
+
+Once that change happens on the server, the clients will start getting
+the correct variant of the resource based on their `version` key.
```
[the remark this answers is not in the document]
markdroth: The change should propagate atomically through each layer.  I've added a paragraph explaining this.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: As you say, that would work in only a subset of cases, so even if we did that, we'd need this best practice anyway.  And once we have this best practice, we don't need that other one.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: That might work for a simple deployment with no caching xDS proxies.  But as soon as you introduce caching proxies, you need to have the same ordering/priority logic in both the authoritative server and in all of the caching proxies, or else you'd get very non-deterministic results.  I don't think that would wind up being feasible.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: I've added a section above describing implementation considerations that illustrates a naive implementation and suggests an approach for optimization.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
fuqianggao: Actually, we can't get rid of this. Because there are two levels. This MatchType applies to Constraint. While DynamicParameterConstraints.constraints is repeated ConstaintList. Maybe we should name it to constraint_lists to reduce some confusion?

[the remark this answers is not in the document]
markdroth: Okay, great, if we agree that we can remove this, then we can simplify a bit.  Thanks!

Before I make any changes, I'll wait for more feedback from @htuch and @mattklein123 on the thread below about how extensible we want to make this, given the implications for cache performance and difficulty of making changes later.  I think that once we have consensus on where to land on this spectrum, it should be easier to spell out the details.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,692 @@
...
+clients before the corresponding constraints are added on the resources,
+which we expect to be the common case.  (In general, we expect clients
+to send a lot of keys that may not actually be used by the server, since
+deployments often divide their clients into categories before they have
+a need to differentiate the configs for those categories.)
+
+As mentioned above, this approach does introduce the possibility of
+matching ambiguity in certain cases, where there may be more than one
+variant of a resource that matches the dynamic parameters specified by
+the client.  If an xDS transport protocol implementation does encounter
```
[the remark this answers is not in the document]
ejona86: > This would cause the client to constantly flip back and forth between the two variants, depending on which one was most recently updated.

I was making the point that the transport protocol does not have ambiguity here, because if the server is alternating on which resource it sends, the client is only entitled to operate in one way. The server should get its act together, sure, but clients and proxies are ignorant to this sort of issue.

> With regard to deletion, note that the constraints on a resource are part of its cache key and therefore part of its unique identity in the context of the transport protocol.

This proposal does not provide a way to delete a specifically-constrained resource in the incremental protocol. You can only delete by resource name (`repeated string removed_resources = 6;`).

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
ejona86: > So I think that the best way to promote interoperability is to have clients do this check anyway, to make sure that servers are not just forgetting to send the constraints.

"forgetting to send the constraints" is pretty low on my list of concerns. If a server doesn't want the benefits of a proxy it seems like that's its prerogative and not much we can do about it. After all, it could continue using Node metadata or send empty constraints.

Similarly, you can't really "make" clients do the check, unless failing to do so would change behavior. And right now, it doesn't change the behavior. But we can say that clients are encouraged to do the check (and then we implement it in Envoy and gRPC).

The biggest piece I think is useful is to realize the check should be easier in end-clients because it isn't "select the correct subscription" but instead "does the returned resource match the client's constraints." That style sounds much less onerous/bug-prone to implement.

The biggest server bug risk seems to be _partial_ constraints, where the constraints don't fully encompass the selection logic. And I think you need a proxy before you'll find those. Non-matching constraints, by contrast, is the easy bug to notice; clients checking constraints only helps the case that is already easy.

It should be noted that a reason not to do the check in clients is to reduce ossification of the constraints. If there's a new constraint type, you can coordinate support in just servers and proxies. Although there are few client implementations, upgrading to newer versions may be a slower process and prevent some server-side optimization.

> because if the constraints are empty, they will still match.

That sounds okay to me. The server isn't participating and thus doesn't get the benefits, and it may not even be using constraints. I really don't see any value in requiring constraints to be present.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,946 @@
...
+
+### Dynamic Parameters
+
+With the above property in mind, this document proposes the following
+data structures:
+- **Dynamic parameters**, which are a set of key/value pairs sent by the
+  client when subscribing to a resource.
+- **Dynamic parameter constraints**, which are a set of criteria that
+  can be used to determine whether a set of dynamic parameters matches
+  the constraints.  These constraints are part of the cache key for an
```
htuch, line 112: Is it correct to say that this is part of the cache key or rather that the constraints are used during the cache lookup process? The distinction is that a key is usually a pure transformation of some input to a key space, here we are doing something more like constraint satisfaction.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
htuch: I think it probably makes sense to version the transport protocol itself to support this future proofing. It can be orthogonal to this xRFC but maybe we can add it at the same time as the use of this?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
htuch: I think flat is OK actually; any tree can be transformed to https://en.wikipedia.org/wiki/Disjunctive_normal_form.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
htuch: Reading further, I think the ambiguity is on what the cache key is. Is this a cache keyed by the resource names / constraints provided by the server, or a cache for the client, keyed by the dynamic parameters provided by the client? It might be worth clarifying this in various places to reduce reader potential for confusion.

htuch (commented):
LGTM modulo minor comments on writeup.

mattklein123:
@markdroth sorry catching up from leave. Where are we at with this? Should I do another pass and then we can get this merged or are there other pending changes I should wait for?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
markdroth: > This proposal does not provide a way to delete a specifically-constrained resource in the incremental protocol. You can only delete by resource name (`repeated string removed_resources = 6;`).

Good catch!  I've just pushed a change to fix this.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: @mattklein123 This is the main open issue that I'd like your feedback on.  I think we need to make some trade-offs here between flexibility and complexity, but we may want to err a bit on the side of more flexibility, given that this will be hard to change later.

@htuch I agree that we could address this by versioning the transport protocol.  We've talked before about possibly divorcing the transport protocol versioning from the data model versioning, and I do think that's something we should do.  But we'll need to think carefully about how to do this, and I think it will have to be a separate design.  I don't think we actually need to do it until we actually need to make a change to whatever we put in this proposal, though.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
markdroth: The intent here is that the cache key is both the resource name and the constraints.  It has to be that way, because we need to cache each variant of the resource separately, and each variant is uniquely identified by its name and constraints.  See the "Considerations for Implementations" section below for more details.

Do you have any specific suggestions as to changes I could make to clarify this?

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
markdroth: @mattklein123 This thread is also related to the other one that I tagged you on.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
htuch: Can you add an explicit definition of cache key somewhere?

markdroth: I've changed the wording to avoid the term "cache key".  Does that address your concern?

htuch: Yep, thanks.

htuch:
LGTM, will defer to @mattklein123 for final approval/merge.

mattklein123 (commented):
Sorry for the delay. In general this looks great. My general feeling though is we should plan for the expansion of matching capabilities. Personally, I like the idea of client capabilities, as that would allow us to not only handle additions, but also potentially features that certain implementations don't want to support (e.g., "crazy regex matching option").

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,973 @@
...
+    // A list of key/value constraints.
+    repeated Constraint constraints = 1;
+
+    // How to match the constraints.
+    MatchType match_type = 2;
+  }
+
+  // A list of constraint lists. All constraint lists must match (i.e.,
+  // logical AND semantics).
+  repeated ConstraintList constraints = 1;
```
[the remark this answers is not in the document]
markdroth: Done.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -0,0 +1,973 @@
...
+This section documents limitations and design alternatives that we
+considered.
+
+### Limitation on Enhancing Matching in the Future
+
+One limitation of this design is that, because all xDS transport protocol
+implementations (clients, servers, and caching proxies) need to implement
+this matching behavior, it will be very difficult to add new matching
+behavior in the future.  Doing so will probably require some sort of
+client capability.
```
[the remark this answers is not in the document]
markdroth: I think we can add the client capabilities when the time comes.  It's not clear to me that we gain anything by doing it now, since they won't provide any benefit right now.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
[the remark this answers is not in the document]
markdroth: @fuqianggao Please take a look at the latest changes and verify that they should be implementable for TD.

markdroth (commented):
I've updated the doc as per our recent discussions.  PTAL.

markdroth:
I've sent envoyproxy/envoy#20926 with the proto changes.

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
```diff
@@ -562,25 +515,15 @@ that a server has the following two variants of a resource:
```
fuqianggao, line 515: Are you saying that this type of configurations are not allowed based on the best practice?

markdroth: Yes.  If the server always ensures that all variants of a resource have non-overlapping constraints for the same set of keys, then this kind of conflict cannot happen.

fuqianggao: OK. Can you make this a bit more clear?

mattklein123 (approved):
LGTM, thanks for all the back and forth!

htuch (approved):
LGTM, thanks!

proposals/TP2-dynamically-generated-cacheable-xds-resources.md
fuqianggao: Looks good.

Issue #2, opened 2021-03-03
Stable release of Go generated protos


dfawley:
Anyone writing a Go application that wishes to use Envoy protos will need a stable release of this repo.  In case you are not aware, proto definitions cannot be generated independently by libraries, since multiple versions of the generated code for a single message that make their way into the same binary will cause a panic at init time.  And without a stability guarantee, use of this repo as the official source of the protos is risky.  Please create a stable v1.x release of this repo ASAP.  Thank you.

Related to https://github.com/cncf/udpa/issues/42 and https://github.com/envoyproxy/go-control-plane/issues/391.

cc @htuch

htuch:
@dfawley what do we need to do to make this "stable"? Is it just a release tag that is v1.0.0? I'm not sure why Go doesn't work with < v1.0.0 here, but we can probably do a v1.0.0 and iterate.

https://github.com/cncf/xds is now live and the canonical xDS-WG repo.

dfawley:
> @dfawley what do we need to do to make this "stable"? Is it just a release tag that is v1.0.0?

Yes, exactly!

> I'm not sure why Go doesn't work with < v1.0.0 here, but we can probably do a v1.0.0 and iterate.

Everything _technically_ works fine without a v1 release.  This is purely symbolic -- releasing v1 is a signal that the APIs are stable, and that the owners will not break backward compatibility.

phlax:
not sure if i follow entirely - but im wondering if a release is created should it not be the version number of the api - ie `v3.x`

phlax:
nm - ignore me - this is xds version not envoy's api version

dfawley:
Exactly.  This repo can be at v1 and could eventually have envoy/xds v4-v7 protos in it as well as v3.

htuch:
@markdroth @mattklein123 @lizan are you happy with cutting v1 from this repo? If so I can push the button.

markdroth:
Assuming that we're really treating this just as a checkpoint tag for build-dependency purposes, I think this is fine.  Before we would do any sort of really public release, I'd want to finish renaming the protos in the udpa tree to the xds tree.  But that seems like not an issue here, so let's not block on it.

mattklein123:
I don't have a strong opinion either, as long as per @markdroth we aren't locking ourselves into anything we can't change. If it's an arbitrary point in time for "reasons" that's fine with me.

howardjohn:
[...]

PR #11, opened 2021-09-07
annotations: add wip for messages and fields

mattklein123:
Part of https://github.com/envoyproxy/envoy/issues/17920

Signed-off-by: Matt Klein <mklein@lyft.com>


udpa/annotations/status.proto
```diff
@@ -10,6 +10,15 @@ extend google.protobuf.FileOptions {
   StatusAnnotation file_status = 222707719;
 }
 
+extend google.protobuf.MessageOptions {
```
markdroth, line 13: Instead of adding new annotations in the udpa directory, how about adding these in the xds directory, as per the agreed direction discussed in #2?

markdroth (commented):
One quick change suggestion, but otherwise looks good.

udpa/annotations/status.proto
mattklein123: Sure will do.

htuch (commented):
LGTM modulo the s/udpa/xds/.

mattklein123:
@htuch @markdroth this should be ready to go, thanks.

htuch (approved):
LGTM, thanks!

PR #12, opened 2021-09-22
bazel: update configuration

keith:
This updates bazel itself, as well as the bazel rules, and golang. These
updates are mostly tied together since rule compatibility that far back
isn't solid. This brings this repo closer to what envoy itself actually
has.

Signed-off-by: Keith Smiley <keithbsmiley@gmail.com>


keith:
looks like i have some more work to do here

keith:
@htuch can you check this out when you have a chance?

htuch (approved):
Thanks for the cleanup / modernization of the Bazel rules!

PR #13, opened 2021-09-23
Add go_package to annotations protos

keith:
With newer versions of rules_go and bazel this is required


keith:
@htuch can you check this out when you have a chance?

udpa/annotations/migrate.proto
```diff
@@ -4,6 +4,8 @@ package udpa.annotations;
 
 import "google/protobuf/descriptor.proto";
 
+option go_package = "github.com/cncf/xds/go/annotations";
```
htuch, line 7: Looks good, just curious about path structure; we have both `udpa/annotations` and `xds/annotations/v3`, can we support this sensibly?

keith: I don't fully understand how this all works, but I think in that case you get some full qualification like `xds.annotations.v3`?

htuch: I guess I'm asking is do they share a common package structure in Go land?

keith: I don't think they will, and I don't think they did before. We might be able to try and use the same values here, but I guess we'd have potential for conflicts, not sure how likely that would be?

htuch (approved):

PR #14, opened 2021-10-01
protos: Remove unused imports

phlax:


phlax:
...i need to rem how to regenerate the `.pb.go` file...

PR #15, opened 2021-10-02
Add go_package annotations to all protos

keith:


xds/annotations/v3/status.proto
```diff
@@ -4,6 +4,8 @@ package xds.annotations.v3;
 
 import "google/protobuf/descriptor.proto";
 
+option go_package = "github.com/cncf/xds/go/xds/annotations/v3";
```
keith, line 7: I'm not actually sure if this should include the second xds or not, in the other case I omitted this, but can't recall why

keith:
I guess at this point I should add these to all protos?

htuch:
@keith if you can do them all and use consistent package namespace, SG.

keith:
Ok I got them all here.

htuch (approved):

PR #17, opened 2021-10-07
Duplicate all protos from udpa tree to xds tree

markdroth:
This is step 1 of the plan from #2.

Question: How do I test something like this to make sure that it's not going to break any callers?

### DIFF
--- a/udpa/README.md
+++ b/udpa/README.md
@@ -1,0 +1,10 @@
+# THESE PROTOS ARE DEPRECATED
+
+We are no longer using the "UDPA" name, and we are moving away from the
+protos in this tree.  Users should prefer the corresponding protos in
+the xds tree instead.
+
+No new changes will be accepted for protos in this tree.  If you want to
+change something, make the change to the corresponding proto in the xds
+tree instead, and then change your protos to use the xds proto instead
+of the udpa proto.
--- a/udpa/annotations/migrate.proto
+++ b/udpa/annotations/migrate.proto
@@ -1,3 +1,7 @@
+// THIS FILE IS DEPRECATED
+// Users should instead use the corresponding proto in the xds tree.
+// No new changes will be accepted here.
+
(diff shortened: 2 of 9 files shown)


bazel/api_build_system.bzl
```diff
@@ -146,14 +146,25 @@ def udpa_proto_package(srcs = [], deps = [], has_services = False, visibility =
...
 
-def udpa_go_test(name, **kwargs):
+def xds_go_test(name, **kwargs):
     go_test(
         name = name,
         **kwargs
     )
+
+# Old names for backward compatibility.
+# TODO(roth): Remove these once all callers are migrated to the new names.
```
htuch, line 162: This will be removed in this PR?

xds/service/orca/v3/orca.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
```
htuch, line 1: I assume you'll do another pass where you flag in bold letters that the old UDPA version is frozen, stale etc?

htuch (commented):
Who do you want to test for breakage?

bazel/api_build_system.bzl
markdroth: I was thinking that these build rules might be pulled in by projects that use this repo, so I didn't know if I could remove the old names without breaking anyone.  But if we know that no one is using these rules, then I'm happy to remove them immediately.

xds/service/orca/v3/orca.proto
markdroth: Done.

markdroth (commented):
I was thinking about tests for projects that depend on this.  I can obviously figure out whether this will break gRPC, but, for example, how do I know if this is going to be okay for Envoy?

htuch (approved):

PR #18, opened 2021-10-20
go: Update Go version to 1.17

110y:
- Just execute `go mod tidy -go=1.17` under the `go` directory.
    - Go versions older than 1.16 are not supported version.

### DIFF
--- a/go/go.mod
+++ b/go/go.mod
@@ -1,9 +1,17 @@
 module github.com/cncf/xds/go
 
-go 1.11
+go 1.17
 
 require (
 	github.com/envoyproxy/protoc-gen-validate v0.1.0
-	github.com/golang/protobuf v1.3.2
+	github.com/golang/protobuf v1.5.0
 	google.golang.org/grpc v1.25.1
+	google.golang.org/protobuf v1.27.1
+)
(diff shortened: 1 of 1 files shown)


go/go.mod
```diff
@@ -1,9 +1,17 @@
...
 require (
 	github.com/envoyproxy/protoc-gen-validate v0.1.0
-	github.com/golang/protobuf v1.3.2
+	github.com/golang/protobuf v1.5.0
 	google.golang.org/grpc v1.25.1
+	google.golang.org/protobuf v1.27.1
+)
+
+require (
+	golang.org/x/net v0.0.0-20190311183353-d8887717615a // indirect
```
110y, line 13: https://golang.org/doc/go1.17#go-command

> For the go command to correctly resolve transitive imports using the pruned module graph, the go.mod file for each module needs to include more detail about the transitive dependencies relevant to that module. If a module specifies go 1.17 or higher in its go.mod file, its go.mod file now contains an explicit require directive for every module that provides a transitively-imported package. (In previous versions, the go.mod file typically only included explicit requirements for directly-imported packages.)

PR #24, opened 2021-12-22
matcher: add server name matcher

kyessenov:
This is intended to generalize SNI matching in TLS in Envoy's listener unified matcher implementation.
Signed-off-by: Kuat Yessenov <kuat@google.com>


kyessenov:
cc @snowp @markdroth

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,51 @@
...
+  message DomainMatcher {
+    // A non-empty set of wildcard domain names, e.g. ``www.example.com`` or
+    // ``*.com``.
+    repeated string domains = 1 [ (validate.rules).repeated = {min_items : 1} ];
+
+    // Match action to apply when the server name matches any of the wildcard
+    // domain names.
+    Matcher.OnMatch on_match = 2;
+
+    // Indicates whether this match option should be considered when there is a
```
markdroth, line 38: Are there actually cases in which we would *not* want the most-specific match?

markdroth (commented):
Overall, this looks fine to me, but I'd like input from @snowp as well.

xds/type/matcher/v3/domain.proto
kyessenov: For example, let's say we want to have 3 cases:
* `mail.google.com` - most specific match, may have some additional condition in sub-tree.
* `*.google.com` - do some redirect, should match `mail.google.com` if the nested condition fails.
* `*` - drop.

Having one default case can be handled with `on_no_match`. Having two default cases requires two levels of `on_no_match` which means some conditions have to be duplicated twice.

markdroth: I think I see what you want here, but the way it's expressed isn't very intuitive.  Instead, how about calling this option something like `ignore_if_nested_matcher_fails`?  That way, you can set it to true on the entry for `mail.google.com`, which will indicate that that entry will be ignored if the nested matcher fails, in which case it will fall back to `*.google.com` as the next-most-specific match.

kyessenov: The name followed the pattern in https://github.com/cncf/xds/blob/main/xds/type/matcher/v3/ip.proto#L48, but I agree it is not super intuitive. What I want is actually structured fallback, so the bit should be set on `*.google.com` instead.

I realize this matcher is probably a candidate for virtual host matching in Envoy. @snowp is there some prior art to express trie-matching for wildcard domains?

kyessenov: Dropped this field until we need it.

kyessenov:
@snowp Please take a look. I wonder what your thoughts are on compatibility with https://github.com/envoyproxy/envoy/blob/3aacf2bee32fe77f9eee3ffc668fc7942ddd1418/api/envoy/config/route/v3/route_components.proto#L69.

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,46 @@
...
+
+    // Match action to apply when the server name matches any of the wildcard
+    // domain names.
+    Matcher.OnMatch on_match = 2;
+  }
+
+  // Match a server name by multiple wildcard domain matchers. If there are
+  // multiple wildcards matching the server name, the order of declaration is
+  // used to select the first match.
+  repeated DomainMatcher domain_matchers = 1;
```
snowp, line 45: Would it be possible to reduce the nesting here and leverage the ListMatcher to express this "first match wins"?

snowp (commented):
Seems good to me for what it tries to do, just one question on whether we can simplify

xds/type/matcher/v3/domain.proto
kyessenov: It works in tandem with the longest suffix so I guess not. E.g. if we have two matchers for `*.com` but also have `example.com` that takes priority.

snowp: Ok so there can be overlapping domains coming from each entry in this list? At first glance it looked like a list of independent suffixes to match on, so maybe the docs can be clearer?

kyessenov: On a second though, let's just make all wildcards distinct across all matchers. WDYT?

snowp (dismissed):
LGTM, thanks!

kyessenov:
Gentle ping @markdroth .

kyessenov:
@markdroth This is one bit of lost functionality. Can you approve this so we can implement it in Envoy?

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_multiple_files = true;
+option go_package = "github.com/cncf/xds/go/xds/type/matcher/v3";
+
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Server name matcher]
+
+// Matches a fully qualified server name against a set of wildcard domain
+// names.
+message ServerNameMatcher {
```
howardjohn, line 21: Upfront, I completely agree we need something here to get parity with current Envoy config, and I have no problem with this being that solution.

Just wondering if we want a more general vs domain specific matcher. It seems like this could be a "PrefixOrExactMatcher" or something generic, where if there is a `*` we treat it as prefix, else exact?

I think the only behavioral implication would be that `*w.example.com` would be valid. I don't have any concerns with that though; its up to the control plane to reject that if it doesn't like it?

Just an idea, I don't have a strong opinion here

xds/type/matcher/v3/domain.proto
howardjohn: Could you put all the exact matchs first (any order), then the wildcard matches (sorted by length) to get the same behavior?

kyessenov: The key is that the implementation should be non-linear. Lists require traversal which doesn't scale well. This one can be done with an efficient compressed trie.

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,44 @@
...
+    // ``*.com``, or ``*``.
+    repeated string domains = 1 [ (validate.rules).repeated = {min_items : 1} ];
+
+    // Match action to apply when the server name matches any of the wildcard
+    // domain names.
+    Matcher.OnMatch on_match = 2;
+  }
+
+  // Match a server name by multiple wildcard domain matchers. Each wildcard
+  // domain must appear at most once across the domain matchers.
```
mattklein123, line 42: Can you clarify what happens if a non-wildcard domain is specified? Is that allows per above or will it be rejected?

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,44 @@
...
+  // wildcard symbol ``*`` must appear at most once as the left-most part of
+  // the domain. The wildcard does not match an empty string.
+  //
+  // The server name will be matched against all wildcard domains starting from
+  // the longest suffix, i.e. ``www.example.com`` input will be first matched
+  // against ``www.example.com``, then ``*.example.com``, then ``*.com``, then
+  // ``*``. Note that partial wildcards are not supported, and values like
+  // ``*w.example.com`` are invalid.
+  message DomainMatcher {
+    // A non-empty set of wildcard domain names, e.g. ``www.example.com``,
```
mattklein123, line 32: www.example.com isn't wildcard, so a little confusing, and related to my other comments.

mattklein123 (changes requested):
Thanks LGTM with small comments

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,44 @@
...
+option java_package = "com.github.xds.type.matcher.v3";
+option java_outer_classname = "ServerNameMatcherProto";
+option java_multiple_files = true;
+option go_package = "github.com/cncf/xds/go/xds/type/matcher/v3";
+
+option (xds.annotations.v3.file_status).work_in_progress = true;
+
+// [#protodoc-title: Server name matcher]
+
+// Matches a fully qualified server name against a set of wildcard domain
```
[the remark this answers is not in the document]
kyessenov: Yes, exact names are allowed. The wildcard, if present, must be the leftmost symbol before the dot.

xds/type/matcher/v3/domain.proto
kyessenov: It's allowed. I spelled out that exact are allowed.

kyessenov:
Thanks, updated the documentation to make it easier to follow.

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,46 @@
...
+
+// Matches a fully qualified server name against a set of domain
+// names with optional wildcards.
+message ServerNameMatcher {
+  // Specifies a set of exact and wildcard domains and a match action. The
+  // wildcard symbol ``*`` must appear at most once as the left-most part of
+  // the domain on a dot border. The wildcard matches one or more non-empty
+  // domain parts.
+  message DomainMatcher {
+    // A non-empty set of domain names with wildcards, e.g.
```
mattklein123, line 27:
```suggestion
    // A non-empty set of domain names with optional wildcards, e.g.
```

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,46 @@
...
+    // ``www.example.com``, ``*.com``, or ``*``.
+    repeated string domains = 1 [ (validate.rules).repeated = {min_items : 1} ];
+
+    // Match action to apply when the server name matches any of the wildcard
```
mattklein123, line 31:
```suggestion
    // Match action to apply when the server name matches any of the
```

xds/type/matcher/v3/domain.proto
```diff
@@ -0,0 +1,46 @@
...
+    Matcher.OnMatch on_match = 2;
+  }
+
+  // Match a server name by multiple domain matchers. Each domain, exact or
+  // wildcard, must appear at most once across all the domain matchers.
+  //
+  // The server name will be matched against all wildcard domains starting from
+  // the longest suffix, i.e. ``www.example.com`` input will be first matched
+  // against ``www.example.com``, then ``*.example.com``, then ``*.com``, then
+  // ``*``, until the associated matcher action accepts the input. Note that
```
mattklein123, line 42: What is a partial wildcard? Clarify?

mattklein123 (changes requested):
Thanks LGTM with small comments

markdroth:
Sorry for the delay here.

LGTM, modulo comments from @mattklein123.

xds/type/matcher/v3/domain.proto
kyessenov: Ack.

xds/type/matcher/v3/domain.proto
kyessenov: Re-worded, thanks. This was straight copied from envoy originally.

mattklein123 (approved):
Thanks!

PR #28, opened 2022-02-23
matchers: add action name as opposed to full configuration

kyessenov:
This is useful to reduce verbosity for string-value actions, e.g. picking a listener by name using the unified matchers.

Signed-off-by: Kuat Yessenov <kuat@google.com>


kyessenov:
CC @snowp

kyessenov:
Yes, there is too much boilerplate when using string-valued actions (see the linked PR). Open to other ways but this looked least disruptive.

PR #31, opened 2022-03-08
matchers: add CEL matcher, input

sergiitk:
Draft. I have trouble building it because of `@com_google_googleapis` dependencies. I had partial success building, but I had to update googleapis repo, and/or modify `api_build_system.bzl`. Before I spend more time on it, please take a look.

@markdroth @kyessenov @yanavlasov


xds/type/matcher/v3/http_inputs.proto
```diff
@@ -0,0 +1,51 @@
...
+
+// [#protodoc-title: Common HTTP Inputs]
+
+// Match input indicates that matching is done on the set of :ref:`HTTP attributes
+// <https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes>`.
+// The attributes will be exposed via Common Expression Language runtime to associated CEL matcher,
+// or via ``get_property`` ABI method to associated Wasm extension based matcher.
+message HttpCelMatchInput {
+  // [#next-free-field: 7]
+  message Attributes {
```
kyessenov, line 26: Can you explain the intent behind limiting visibility? Is it to "hide" subset of attributes and for what purpose?

I'd imagine that the HTTP input would try to provide all information to the extent possible and leave some values undefined if they are not available. E.g. response would be empty on request headers.

kyessenov: CEL distinguishes between empty and missing values. It's an error if some top-level property is not defined which propagates as the evaluation result.

I think the presence of top-level attributes is pretty clear depending on the context. So the new capability is that one might want to hide some attributes when they are available. Is that capability really needed?

sergiitk: Updated `HttpCelMatchInput` in https://github.com/envoyproxy/envoy/commit/8c20ec1384444fefd433a3b839ee0bc2f78bfed5. Now looking at `CelMatcher`.

sergiitk:
Re: `skip_if_error` discussed IRL. Decided for this particular purpose of using in unified matchers, makes more sense to just clarify it in the comment.

sergiitk:
@markdroth @yanavlasov @tyxia Please take a look. I'm yet to fix the build (the problem with including googleapis dependencies need to be solved), but the protos are ready.

sergiitk:
The doc will look something like this:

image

xds/type/matcher/v3/cel.proto
```diff
@@ -0,0 +1,60 @@
...
+//   In all other cases, the match is ``false``, including but not limited to: non-bool types,
+//   ``false``, ``null``,`` int(1)``, etc.
+//   In case CEL expression raises an error, the result of the evaluation is interpreted "no match".
+//
+// Refer to :ref:`Unified Matcher API <envoy_v3_api_msg_.xds.type.matcher.v3.Matcher>` documentation
+// for usage details.
+//
+// [#comment:TODO(sergiitk): Link HttpAttributesMatchInput + usage example.]
+// [#comment:TODO(sergiitk): Add extension category.]
+message CelBoolMatcher {
```
markdroth, line 37: I think this can just be called `CelMatcher`.

xds/type/matcher/v3/cel.proto
```diff
@@ -0,0 +1,60 @@
...
+// <https://github.com/google/cel-spec>`_ (CEL) expression against the standardized set of
+// :ref:`HTTP attributes <arch_overview_attributes>`.
+//
+// .. attention::
+//
+//   Besides CEL evaluation raising an error explicitly, CEL program returning a type other than
+//   the ``string``, or not returning anything, are considered an error as well.
+//
+// [#comment:TODO(sergiitk): Add extension category.]
+message CelExtractString {
```
markdroth, line 52: This isn't a matcher, so it doesn't really belong in this file.  Suggest moving it to xds/type/v3/cel.proto instead.

markdroth (commented):
Overall structure looks good!  Just a couple of organizational comments.

xds/type/matcher/v3/cel.proto
sergiitk: Makes sense. I was looking at [`envoy.type.matcher.v3.RegexMatchAndSubstitute`](https://github.com/envoyproxy/envoy/blob/003947cd2f77476f67849ce4b8f710cc034c024e/api/envoy/type/matcher/v3/regex.proto#L72) (which also isn't a matcher, but lives in matchers) but didn't notice it has been removed in `xds.type.matcher.v3` / [regex.proto](https://github.com/cncf/xds/blob/main/xds/type/matcher/v3/regex.proto).

sergiitk:
Thanks for the quick review, @markdroth. Addressed your comments.

markdroth:
For future reference, please don't force-push once a review has started; that makes it harder for the reviewer to see what's changed.  If you need to merge master, just merge master into your branch and push that merge commit.

markdroth:
This looks good to me!  Just need to fix CI.

sergiitk:
>  For future reference, please don't force-push once a review has started; that makes it harder for the reviewer to see what's changed. If you need to merge master, just merge master into your branch and push that merge commit.

My apologies, normally I reserve to this in exceptional circumstances. This time I have accidentally overridden `AuthorDate` in all commits, when was adding missing `--signoff`.

sergiitk:
Force-pushed it again while setting up automatic signoff :( Sorry

sergiitk:
@kyessenov @markdroth ready for re-review. The build is fixed.

Note: there's a couple of TODOs. I'm planning to address them in a follow up PR, once envoy picks up this change, and doc generation works.

markdroth (approved):
Looks great!

PR #40, opened 2022-06-09
Add pattern to string.proto

silverstar194:


adisuissa:
cc @adisuissa

xds/type/matcher/v3/string.proto
```diff
@@ -52,6 +52,14 @@ message StringMatcher {
     //
     // * *abc* matches the value *xyz.abc.def*
     string contains = 7 [(validate.rules).string = {min_len: 1}];
+
+    // The input string must have the pattern specified here.
+    // Note: empty pattern match is not allowed
+    //
+    // Examples:
+    //
+    // abc/{lang}/{county} matches the value abc/us/eng
```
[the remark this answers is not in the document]
adisuissa: +1.
Can you clarify where `lang`/`county` categories and their possible values are defined?

xds/type/matcher/v3/string.proto
```diff
@@ -52,6 +52,23 @@ message StringMatcher {
...
+    //
+    // Note: There is overlap between regex and pattern.
+    // Pattern serves as a better user experience through simplified matching syntax.
+    //
+    // Examples:
+    //
+    // abc/{lang}/{county} matches the value abc/eng/us
+    // Note format of: abs/<user-defined-var>/<user-defined-var>
+    // {lang} == "eng" and {country} == "us"
+    string pattern = 8 [(validate.rules).string = {min_len: 1}];
```
adisuissa, line 71: Sorry I think I'm missing something.
Is the idea to match a string given some pattern, and assign the values to their variables (e.g., lang will be "eng"), or are there known categories and known possible values for these categories, and a match will occur if the used values are in the categories.

silverstar194: The will be matched and assigned to variables. No know categories.

adisuissa: How would the string "abcdef" match against pattern="abc{first}{second}", as the possible assignments are possible:
first="de", second="f"
first="d", second="ef"
?

silverstar194: This would not be a valid config. Only items between "/"s are used as vars.

adisuissa: This is an interesting use-case, and I'm not sure if it should be specified as part of the StringMatcher (the StringMatcher just matches according to some pattern, and doesn't assign values to external variables).

@snowp was this kind of use-case ever considered?

silverstar194: The variable extraction will be done elsewhere. Let's just keep the matcher a matcher

silverstar194: I've updated comments/docs in code to reflect this.

xds/type/matcher/v3/string.proto
```diff
@@ -52,6 +52,25 @@ message StringMatcher {
...
+    // Each value can be a string literal or wildcard (*).
+    //
+    // Empty pattern match is not allowed.
+    //
+    // Note: There is overlap between regex and pattern.
+    // Pattern serves as a better user experience through simplified matching syntax.
+    //
+    // Examples:
+    //
+    // abc/*/* matches the value abc/eng/us
```
adisuissa, line 67: Each example should be in bullet (line prefix with `*`). Also each of the actual pattern/value should be italic ([reference](https://thomas-cokelaer.info/tutorials/sphinx/rest_syntax.html#inline-markup-and-special-characters-e-g-bold-italic-verbatim)). See rendered output of previous docs: https://www.envoyproxy.io/docs/envoy/latest/api-v3/type/matcher/v3/string.proto

Unfortunately I don't think the cncf repo runs the docs rendering script in a PR. @phlax is this correct? How do we make sure these docs are rendered well?

xds/type/matcher/v3/string.proto
```diff
@@ -52,6 +52,25 @@ message StringMatcher {
...
+    // Note format of: abs/<wildcard-1>/<wildcard-2>
+    // {wildcard-1} == "eng" and {wildcard-2} == "us"
```
adisuissa, line 69: I think that wildcard-1/2 makes this more confusing, as the matcher only matches the entries.

xds/type/matcher/v3/string.proto
```diff
@@ -52,6 +52,25 @@ message StringMatcher {
...
+    //
+    // abc/foo matches the value abc/foo
```
adisuissa, line 71: Same as above.

xds/type/matcher/v3/string.proto
silverstar194: changed

xds/type/matcher/v3/string.proto
silverstar194: changed

xds/type/matcher/v3/string.proto
markdroth: This type of matching seems very specific to path matching, but StringMatcher can be used for things other than paths, so I'm not sure this is really the right place for this.  I think it would make more sense to create a new matcher for this purpose.

silverstar194: It can be used in any context where simplified wildcard matching based on a delimiter is needed.

silverstar194: I am ok with expanding the / to be generalized

markdroth: I think that even with making this more generalized, it still seems to be a fairly special case.  I think it makes more sense for this to be a separate matcher, just like RegexMatcher is.

xds/type/matcher/v3/string.proto
phlax: envoy pulls in the xds protos as a dep so when the dep is updated the newer protos are rendered

if you need to test a PR ahead of this you can create a PR in envoy and point the dep at the PR (commit hash for PR mostly will work as release version in `bazel/repository_locations.bzl`)

PR #52, opened 2022-11-11
Remove unused proto

HackToHell:
This causes newer version of protoc to fail from bazel.


phlax:
i think this is a duplicate of #43 which also has update go files

HackToHell:
Ah perfect, missed it somehow.

phlax:
ill ping the maintainers and see if we can get that one landed

PR #54, opened 2022-12-15
orca: replace uint64 rps with double rps_fractional

markdroth:
Signed-off-by: Mark D. Roth <roth@google.com>

Need to represent this as a double to avoid rounding error in low-RPS services, which can greatly skew endpoint weights based on this value.

CC @yousukseung


xds/data/orca/v3/orca_load_report.proto
```diff
@@ -21,9 +21,8 @@ message OrcaLoadReport {
   // resources. This should be derived from the latest sample or measurement.
   double mem_utilization = 2 [(validate.rules).double.gte = 0, (validate.rules).double.lte = 1];
 
-  // Total RPS being served by an endpoint. This should cover all services that an endpoint is
-  // responsible for.
-  uint64 rps = 3;
+  // Deprecated -- use qps instead.
```
adisuissa, line 24: nit: keep old comment, and add the deprecation comment.

adisuissa (commented):
LGTM, modulo comment nit.
I don't really understand why double is preferred, especially as other places in the API rps/qps are int-based AFAICT. Maybe add some background details in the field's doc-string or in the PR description.

xds/data/orca/v3/orca_load_report.proto
markdroth: Done.

adisuissa: nit (for the rendered docs):
```suggestion
  // Deprecated -- use ``qps`` instead.
```

adisuissa (approved):
Thanks, left one small nit, but otherwise LGTM.

adisuissa (approved):
LGTM, thanks!
/lgtm api

xds/data/orca/v3/orca_load_report.proto
```diff
@@ -33,4 +34,8 @@ message OrcaLoadReport {
   // available, derived from the latest sample or measurement.
   map<string, double> utilization = 5
       [(validate.rules).map.values.double.gte = 0, (validate.rules).map.values.double.lte = 1];
+
+  // Total QPS being served by an endpoint. This should cover all services that an endpoint is
+  // responsible for.
+  double qps = 6 [(validate.rules).double.gte = 0];
```
htuch, line 40: I'm fine with the move to `double`, that makes sense for stats, but I'm less enthusiastic about switching from RPS to QPS terminology as the way to distinguish the fields. They are synonyms, and a large chunk of the world uses RPS, whereas Google uses QPS and Envoy / xDS has significant influence from this. 

We should pick RPS / QPS as the xDS preferred standard, and put it in a style guide at the very least.

markdroth: As per our offline conversation, I've renamed `qps` to `rps_fractional`.

PR #60, opened 2023-04-26
Allow cpu_utilization larger than 1.0.

yousukseung:
Signed-off-by: Yousuk Seung <ysseung@google.com>


yousukseung:
+ @markdroth @ejona86 @dfawley

yousukseung:
@htuch PTAL.

markdroth:
I think this is the right thing to do, since there are a lot of cluster systems out there that have varying levels of oversubscription and opportunistic provisioning, so I think it will not be uncommon to need values greater than 1.0 here.

xds/data/orca/v3/orca_load_report.proto
```diff
@@ -14,8 +14,9 @@ import "validate/validate.proto";
 
 message OrcaLoadReport {
   // CPU utilization expressed as a fraction of available CPU resources. This
-  // should be derived from the latest sample or measurement.
-  double cpu_utilization = 1 [(validate.rules).double.gte = 0, (validate.rules).double.lte = 1];
+  // should be derived from the latest sample or measurement. The value may be
+  // larger than 1.0 when the usage exceeds the soft limit.
```
htuch, line 18: "exceeds a reporter dependent notion of soft limits". I think it's legit, just want to be clear who is specifying the soft limit.

htuch:
Needs DCO fix (or possibly new PR).

xds/data/orca/v3/orca_load_report.proto
yousukseung: Updated.

yousukseung:
I rebased to fix the DCO error per the link above.

htuch (approved):

PR #61, opened 2023-05-25
cel: add a description to the AST

kyessenov:
This is useful to assist debugging since AST is not easy to de-construct.


xds/type/matcher/v3/cel.proto
```diff
@@ -35,4 +35,7 @@ option (xds.annotations.v3.file_status).work_in_progress = true;
 message CelMatcher {
   // Either parsed or checked representation of the CEL program.
   type.v3.CelExpression expr_match = 1 [(validate.rules).message = {required: true}];
+
+  // Human-readable description of the CEL AST, e.g. the original expression text.
```
adisuissa, line 39: Consider clarifying that this could be used for debugging and is of free form (I'm assuming it is), so no need to adhere to some format.

adisuissa (commented):
LGTM, thanks!
Left a small suggestion, LMK what you think.

xds/type/matcher/v3/cel.proto
kyessenov: Made it a bit more clear about _what_ and _why_.

adisuissa (approved):
LGTM, thanks!

PR #62, opened 2023-06-05
Add application_utilization

yousukseung:
gRPC's client-side weighted round robin currently uses 'cpu_utilization' but the load balancing works with any utilization. Server applications that are not CPU bound can report an application specific utilization metric instead and balance the load better. This change introduces a new top-level field 'application_utilization' for this use.


xds/data/orca/v3/orca_load_report.proto
```diff
@@ -47,4 +47,10 @@ message OrcaLoadReport {
 
   // Application specific opaque metrics.
   map<string, double> named_metrics = 8;
+
+  // Application specific utilization expressed as a fraction of available
```
markdroth, line 51: Instead of saying "Application specifc", I suggest saying "Application-defined".

Also, I think it would be useful to give some examples of how this is expected to be used.  For example, the application might want to look at the utilization of multiple resources and then send the max of those values.

yousukseung: Added an example, PTAL.

But this proto already uses the term "application specific" with `request_cost` and `named_metrics`, I guess it's more consistent to use the same term here?

markdroth:
LGTM.  I'd like to get a non-gRPC review from @htuch.

Also, looks like the DCO check is failing.  @yousukseung, please fix.

htuch (approved):

htuch:
Please fix DCO!

yousukseung:
> Please fix DCO!

Fixed.

PR #63, opened 2023-06-07
status annotation: new status frozen

wbpcode:
Add a new status frozen to the file/message/field to mark a file/message/field is frozen and no any new features or changes should be added to it.

This status is used as a inner tips notify the developers don't change or update the frozen API.

When a API is nearly to be deprecated but we still cannot do it for some reasons, we can mark it as `frozen`.


wbpcode:
cc @markdroth

wbpcode:
This is needn't for now.

PR #64, opened 2023-06-10
Correct a bad path in go_package directives

pugmajere:
There's a missing "udpa" in the go_package path for these protos, which means generated golang code from protos that import these will generate an invalid path.


adisuissa (approved):
LGTM, thanks!

pugmajere:
Do I need to do anything else to get this merged?

PR #69, opened 2023-10-06
Add canonical CEL support: dev.cel.expr (rules_python approach)

sergiitk:

PR #74, opened 2023-11-11
bazel: Migrate py_proto_library to `@com_github_grpc_grpc`

sergiitk:
`py_proto_library` provided by `@com_google_protobuf//:protobuf.bzl` has been deprecated for a while now: 

> This is provided for backwards compatibility only.  Bazel 5.3 will
> introduce support for py_proto_library, which should be used instead.\
> https://github.com/protocolbuffers/protobuf/blob/32af7d211b85f71920acdfa9b796db008f8c2a45/protobuf.bzl#L642-L644

However, native `py_proto_library` has never been  provided, see https://github.com/bazelbuild/bazel/issues/3935. Instead `@rules_python//python:proto.bzl` is [recommended](https://github.com/bazelbuild/bazel/issues/3935#issuecomment-1680850640). I attempted switching to this library, but it's not compatible with `@com_google_googleapis`'s py_proto_library targets, see https://github.com/cncf/xds/pull/69. I found a hacky workaround by using `cc_proto_library` to generate python targets, but downstream integration into Envoy failed (https://github.com/envoyproxy/envoy/pull/30159).

This PR migrates `py_proto_library` implementation to to `@com_github_grpc_grpc`. This implementation is used by  `@com_google_googleapis`'s, and, more importantly, uses bazel aspects. Which decouples cncf/xds and Envoy's dependencies from concrete upstream py_proto_library implementations.

This resulted in a significant code improvement of `bazel/api_build_system.bzl`:
- No more custom `@com_google_googleapis` dependency mapping needed via `py_proto_library` rules override.
- No more hardcoded dependencies `_xds_py_proto_library` - proto dependency tree is determined from `proto_library` definitions via Basel aspects.
- No more `EXTERNAL_PROTO_PY_BAZEL_DEP_MAP` dependency map needed - for the same reason.

Similar work in Envoy: https://github.com/envoyproxy/envoy/pull/30834.


sergiitk:
@adisuissa @kyessenov this is ready for review.
@htuch - FYI this addresses a bunch of TODO's assigned to you.

htuch:
Overall I like this approach a lot. @phlax @moderation are there any concerns here about taking the dependency from https://github.com/grpc/grpc? It's already a build dependency, and these are Bazel rules, so we should still be able to compile Envoy without linking Google gRPC, but just wanted to check.

phlax:
overall looks fine, and appears address a number of issues - not sure iiuc re grpc dep, but from what i do seems ok

adisuissa (commented):
Overall LGTM, thanks for fixing this!
Will it be possible to have the same change in https://github.com/envoyproxy/envoy/tree/main/api/bazel ?

sergiitk:
@adisuissa yes - I have a PR in progress: https://github.com/envoyproxy/envoy/pull/30834.
Was out yesterday, planning to wrap it up today.

moderation:
LGTM

sergiitk:
Looks like to make it work nicely in Envoy, we'd have to merge this first. Otherwise, bazel complains about missing `_genproto` targets (generated by `@com_google_protobuf//:protobuf.bzl`'s `py_proto_library`), because this cncf/xds used rules_override for googleapis before this PR.

```
ERROR: /private/var/tmp/_bazel_sergiitk/e60ee3f6402a793527d720007beb00cb/external/com_github_cncf_xds/xds/annotations/v3/BUILD:5:18: no such target '@com_google_googleapis//google/api:httpbody_py_proto_genproto': target 'httpbody_py_proto_genproto' not declared in package 'google/api' defined by /private/var/tmp/_bazel_sergiitk/e60ee3f6402a793527d720007beb00cb/external/com_google_googleapis/google/api/BUILD.bazel (Tip: use `query "@com_google_googleapis//google/api:*"` to see all the targets in that package) and referenced by '@com_github_cncf_xds//xds/annotations/v3:pkg_py_proto_genproto'
```

It might be possible to solve this in Envoy without merging this, but it'd be a throw-away work. Let's get this one in first.

adisuissa (approved):
Thanks for fixing this!
Seems that the dependency shepherds are ok with the protoc dependency, so I'm going to approve this PR.
LGTM!

alexeagle:
@sergiitk thanks for this example - but per https://groups.google.com/g/grpc-io/c/44OcoVSmeQk?pli=1 it seems like this is using something deprecated. I'd appreciate any pointers to an "officially recommended" way to generate service stubs with Bazel, on behalf of a lot of the community looking for it.

PR #80, opened 2023-11-22
annotations: correct go_package

howardjohn:
This is currently pointing at a non-existant package. Some proto tools just ignore this and make things work, but it should pont to the real go package for maximum compatibiltiy AFAIK.


udpa/annotations/migrate.proto
```diff
@@ -8,7 +8,7 @@ package udpa.annotations;
 
 import "google/protobuf/descriptor.proto";
 
-option go_package = "github.com/cncf/xds/go/annotations";
+option go_package = "github.com/cncf/udpa/go/annotations";
```
adisuissa, line 11: Shouldn't these be `github.com/cncf/xds/go/udpa/annotations`?

howardjohn: Yes, good catch. updated

adisuissa (approved):
LGTM, thanks!

PR #78, opened 2023-11-20
Add `go build` CI job

sergiitk:
One of the recent PRs broke go users: https://github.com/cncf/xds/issues/76. This PR is to address this type of error by adding `cd go && go build ./...` step to the CI.

Major changes:
- The minimal version of `github.com/cncf/xds/go` is now set to `1.19`

Fixes #77.
Closes #67, #18.

PR #75, opened 2023-11-11
cel: Add canonical CEL (dev.cel.expr) fields

sergiitk:
This PR adds canonical CEL (https://github.com/google/cel-spec/tree/master/proto/cel/expr) fields to `xds.type.v3.CelExpression`. Canonical CEL `cel.expr` was created identical to the `google.api.expr.v1alpha1`, but may be extended in a backward-compatible way.

Nuances:
1. The new fields `cel_expr_parsed` and `cel_expr_checked` are added outside of `oneof expr_specifier` per updated policy change: https://github.com/envoyproxy/envoy/pull/30851
2. `option (validate.required) = true` is removed from the `oneof expr_specifier`, so the users may not presume one of the `parsed_expr`, `checked_expr` will be set.

PR #89, opened 2024-03-19
cel: Add canonical CEL (dev.cel.expr) fields

### LINKED ISSUE #76: Get cel.dev/expr error with go get
After improt `github.com/cncf/xds/go/xds/type/v3` , `go get` command get an error

```
go: main imports
        github.com/cncf/xds/go/xds/type/v3 imports
        cel.dev/expr: cannot find module providing package cel.dev/expr
```
`main.go`
```
package main

import (
        "github.com/cncf/xds/go/xds/type/v3"
)
```

`go.mod`

```
module main

go 1.21
```

sergiitk:
This PR adds canonical CEL (https://github.com/google/cel-spec/tree/master/proto/cel/expr) fields to `xds.type.v3.CelExpression`. Canonical CEL `cel.expr` was created identical to the `google.api.expr.v1alpha1`, but may be extended in a backward-compatible way.

Nuances:
1. The new fields `cel_expr_parsed` and `cel_expr_checked` are added outside of `oneof expr_specifier` per updated policy change: https://github.com/envoyproxy/envoy/pull/30851
2. `option (validate.required) = true` is removed from the `oneof expr_specifier`, so the users may not presume one of the `parsed_expr`, `checked_expr` will be set. 

Related:
1. This replaces original PR #75, which was rolled back to address the issue with go build #76. The orginal PR was approved by @markdroth.
2. Go build CI verification added in #78.
3. cel.dev/expr domain is configured to redirect to [google/cel-spec](https://github.com/google/cel-spec), which has `go.mod` and protobuf published in google/cel-spec/pull/342 / v1.15.0.
4. Downstream envoy PR: envoyproxy/envoy#33160. cncf/xds version will be updated once this PR is merged.
    1. Envoy internal tracking: b/282949246.
6. Closes googleapis/google-cloud-go#9031.
7. Fixes #76.


sergiitk:
For posterity, the fix for the go build (besides the domain setup) was implemented in google/cel-spec#342.

sergiitk:
@phlax please review. cc @tyxia.

phlax:
@sergiitk im not a maintainer of this repo

cc @adisuissa

bazel/external_proto_deps.bzl
```diff
@@ -29,10 +29,14 @@ EXTERNAL_PROTO_GO_BAZEL_DEP_MAP = {
     #    go_googleapis in https://github.com/bazelbuild/rules_go/blob/master/go/dependencies.rst#overriding-dependencies
     "@com_google_googleapis//google/api/expr/v1alpha1:checked_proto": "@go_googleapis//google/api/expr/v1alpha1:expr_go_proto",
     "@com_google_googleapis//google/api/expr/v1alpha1:syntax_proto": "@go_googleapis//google/api/expr/v1alpha1:expr_go_proto",
+    "@dev_cel//proto/cel/expr:checked_proto": "@dev_cel//proto/cel/expr:checked_go_proto",
```
adisuissa, line 32: IIUC these are introduced in addition to the com_googleapis variant to avoid introducing a breaking API change. However, as this is a superset of the com_googleapis project, then I think that conceptually it needs to replace it completely (to avoid violating ODR).
I'm a bit concerned what will happen if these two repos diverge. but as we don't have versioning in place, I don't think we can do much about it.
cc @htuch if there are other ways here.

adisuissa (commented):
Thanks for contributing!
I don't have a comment about the way you implemented this (LGTM), but left a general comment highlighting the issue that this brings into the eco-system.

bazel/external_proto_deps.bzl
sergiitk: @adisuissa Canonical CEL (cel.dev) is meant to replace `v1alpha1` CEL from google APIs package. Canonical CEL will always be extended in a backward-compatible way (as protos normally should be), so there isn't a divergence problem. I'm CCing CEL folks here for more context: @l46kok, @TristonianJones.
Also, @tyxia has more context from the Envoy side.

tyxia: From Envoy perspective, `cel.dev` and `v1alpha1` will co-exist. They are guaranteed to be synced by CEL team.

New users in Envoy are recommended to use `cel.dev` while existing users can still use `v1alpha1`

Thus, `v1alpha1` is deprecated but can not be removed.

htuch: I think this makes sense. When the Envoy PR lands, which shoudl be adding implementation level support (not just API import), it might make sense to audit any remaining references to CEL that goes direct to `google.api` package namespace in the API protos in `envoyproxy/envoy`, e.g. RBAC.

sergiitk: Looks like nothing is blocking this PR from being merged, right?

adisuissa (approved):

sergiitk:
Thanks @mmorel-35. AFAIK this repo is not using bazel modules yet. Could you please let CEL folks know about it? They'll likely to be interested in including it into their docs.

krzysztof293 (approved):
Krzysztof293

PR #92, opened 2024-03-22
pgv_plugin_go : Use protoc_gen_validate instead of xds

mmorel-35:
com_envoyproxy_protoc_gen_validate provides the go_proto_compiler with this exact same configuration


adisuissa (approved):
LGTM, thanks!

PR #93, opened 2024-03-22
Bump io_bazel_rules_go -> 0.46.0

mmorel-35:
With the newest version of rules_go, @go_googleapis is not included anymore.
This includes the use of the same version of gazelle an googleapis as rules_go.
It also use golang dependencies as they are use in go.mod weren't used by gazelle.

This is follows https://github.com/envoyproxy/envoy/pull/33063


phlax:
istio repo is here https://github.com/istio/istio - searching on cncf/xds i reckon it shows where the dependencies are set https://github.com/search?q=repo%3Aistio%2Fistio%20cncf%2Fxds&type=code

might be worth trying an update there

adisuissa (approved):
Thanks @mmorel-35. IIUC https://github.com/istio/istio/pull/50562 implies that it works for everything.
LGTM, thanks!

PR #94, opened 2024-04-19
rules_go//proto : use go_grpc_v2 instead of go_grpc

mmorel-35:
go_grpc is deprecated and shall be replaced by go_proto and go_grpc_v2

This also clearly isolate grpc in a dedicated file.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


adisuissa (approved):
LGTM, thanks!

PR #96, opened 2024-05-24
bazel: Add support for bzlmod

keith:
Based on https://github.com/cncf/xds/pull/95


keith:
cc @mmorel-35

phlax:
> Envoy still uses 6.5.0 bazel version. This requires 7.X, it might a blocker ?

i may be wrong - but when i tried to bump envoy -> 7.x it seemed as though it also required a newer llvm version

not sure if updating would block envoy tho

keith:
Oh yea my intent wasn't to force an envoy bump here, this should continue to be backwards compatible and envoy can include it like normal. the issue that there's a transitive dep on rules_jvm_external which requires bazel 7.x

bazel/api_build_system.bzl
```diff
@@ -117,8 +115,6 @@ def xds_proto_package(
         visibility = ["//visibility:public"],
         deps = depset([_go_proto_mapping(dep) for dep in deps] + [
             "@com_envoyproxy_protoc_gen_validate//validate:go_default_library",
-            "@org_golang_google_genproto_googleapis_api//annotations:annotations",
-            "@org_golang_google_genproto_googleapis_rpc//status:status",
```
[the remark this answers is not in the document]
keith: trying here https://github.com/envoyproxy/envoy/pull/34355

.gitignore
```diff
@@ -1 +1,2 @@
 bazel-*
+MODULE.bazel.lock
```
phlax, line 2: @keith raising this here as its come up in other repo - wondering about checking in lock files

@mmorel-35 was saying that its discouraged due to platform differences - im trying to understand how reproducibility is ensured without doing so (and i guess similar with the added security that using hashes provides)

keith: right now the lock files just aren't very stable, and across platforms i think they can differ. i think soon it will be reasonable to check them in. but bazel is otherwise using a similar algorithm to go so the dep resolution is reproducible, with the only exception I know of being if upstream yanks a version it can break (but that's not the worst thing)

phlax: >  if upstream yanks a version it can break

but without hashes surely if either the dep pull was subject to MITM or upstream republished to same version (perhaps the registry doesnt allow that and that is the protection - similar to eg pypi) then you would have no way o knowing

dont want to block, just trying to understand the state of the art wrt bazel deps

keith: yea MITM for sure. It would have to MITM the registry in general.  Theoretically it's not allowed to republish the same version (but technically possible)

phlax (approved):

phlax:
cc @adisuissa for signoff

.gitignore
phlax: hmm yeah - wondering if the registry itself should be signed a la apt etc

adisuissa (approved):

PR #101, opened 2024-08-22
xDS: Remove WIP from unified matcher

tyxia:
unified matcher has been used in production for quite a while (over 1 year at least). We can still develop new features (e.g., add new custom_matcher) but matcher API itself should not be WIP.

I found WIP status  in the doc is confusing and could introduce unnecessary concern . https://www.envoyproxy.io/docs/envoy/latest/xds/type/matcher/v3/matcher.proto#xds-type-matcher-v3-matcher


tyxia:
Add @htuch and @adisuissa as reviewers

PTAL, Thanks

adisuissa (approved):
LGTM! I think this has been used for quite some time, so should be made stable.
Thanks!

PR #105, opened 2024-10-25
Add ORCA specification

AndresGuedez:
Add a standalone ORCA specification under the top-level orca/ directory.

This specification is intended to be continuously updated as the format evolves. 

Signed-off-by: Andres Guedez aguedez@google.com


orca/orca-spec.md
```diff
@@ -0,0 +1,278 @@
...
+  // the load reporting agent will be returned. This provides an opportunity for
+  // the client to selectively obtain a subset of tracked costs.
+  repeated string request_cost_names = 2;
+}
+```
+
+The definition for this service can be found at
+https://github.com/cncf/xds/blob/main/xds/service/orca/v3/orca.proto.
+
+Alternatively, an HTTP endpoint can be exposed by the load reporting server. A GET or HEAD request
```
markdroth, line 219: This paragraph seems to describe a possible future improvement, not something that's part of the existing ORCA spec.  I suggest either removing it or moving it to a "Possible future improvements" section.

orca/orca-spec.md
```diff
@@ -0,0 +1,278 @@
...
+  // This should be derived from the latest sample or measurement.
+  // The value may be larger than 1.0 when the usage exceeds the reporter
+  // dependent notion of soft limits.
+  double application_utilization = 9 [(validate.rules).double.gte = 0];
+}
+```
+
+# Reporting
+
+## Inline Per-Request Format
```
markdroth, line 151: I think we need to provide some guidance about when it's useful to choose per-request vs. OOB reporting.

I think we should recommend that the vast majority of applications use per-request reporting.  OOB reporting should be considered only for traffic patterns that do not result in per-request reports being delivered often enough to stay fresh, such as a client with very sparse traffic or an application whose traffic is mostly long-running streams.

We should also point out that OOB reporting is good for utilization but doesn't make much sense for query cost reporting.

orca/orca-spec.md
```diff
@@ -0,0 +1,278 @@
...
+# ORCA Integration Considerations
+
+## Sidecar
+
+A sidecar, e.g. Envoy or some other in-container proxy process, can inject the inline ORCA header
+at response completion.
+
+![Sidecar Integration Diagram](./orca-spec-sidecar.jpg)
+
+The advantage of a sidecar model is that the application requires no modification as in the example
```
markdroth, line 232: I think this is true only if the sidecar is capable of determining the metrics that the endpoint wishes to report to its clients.  For example, this would probably work for things like CPU or memory utilization, which the sidecar can determine in a fairly generic way.  But it won't work for application-specific things, like custom utilization metrics or query cost metrics -- those will always require cooperation on the part of the application.

orca/orca-spec.md
```diff
@@ -0,0 +1,278 @@
...
+This method will combine the core metrics such as CPU and Memory (computed by the library) and,
+together with the extra application-specific metrics, compute the inline ORCA response header. The
+application will then be responsible for including this in its response.
+
+# Known DPLB Implementations
+
+## gRPC
+TODO: [Repository]()
+
+[Design](https://github.com/grpc/proposal/blob/master/A51-custom-backend-metrics.md)
```
markdroth, line 264: It may also be useful to refer to the following designs as well:

- [gRFC A58: Client-Side Weighted Round Robin](https://github.com/grpc/proposal/blob/master/A58-client-side-weighted-round-robin-lb-policy.md)
- [gRFC A64: xDS LRS Custom Metrics Support](https://github.com/grpc/proposal/blob/master/A64-lrs-custom-metrics.md)
- (pending) [gRFC A85: Changes to xDS LRS Custom Metrics Support](https://github.com/grpc/proposal/pull/454)

orca/orca-spec.md
```diff
@@ -0,0 +1,278 @@
...
+```
+
+The definition for this service can be found at
+https://github.com/cncf/xds/blob/main/xds/service/orca/v3/orca.proto.
+
+Alternatively, an HTTP endpoint can be exposed by the load reporting server. A GET or HEAD request
+to this endpoint will return a response with the `endpoint-load-metrics` header that contains the
+load report in one of the formats listed above.
+
+# ORCA Integration Considerations
```
markdroth, line 223: I think this section also needs to be updated, since it no longer seems that relevant as currently written.  In general, I think users are not going to choose a sidecar or library model based solely on ORCA; instead, they will probably have one of those models already and will want to know what the options are for adding ORCA support.

I think what we should be explaining here is that there are cases where a smart data plane like Envoy or gRPC can handle ORCA for you, transparently to the application, and there are other cases where the application will need to generate the ORCA data directly.

markdroth (commented):
Thanks for updating this!

orca/orca-spec.md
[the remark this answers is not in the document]
AndresGuedez: Done.

PR #107, opened 2024-10-25
xRFC TP3: Error Propagation

anicr7:
Proposal for the xDS error propagation which allows xDS management servers to provide additional information to the clients in case of errors like permission errors or resource being missing.

cc: @markdroth, @adisuissa, @htuch

### DIFF
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -11,18 +11,13 @@ This proposal introduces enhancements to the [xDS transport protocol](https://ww
 
 The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
 
-This proposal includes a new field for each subscribed Resource, called ResourceError. This field will provide detailed information resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the user. 
-
+This proposal includes a new field for each subscribed Resource, called `ResourceError`. This field will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the user. 
 
 ## Background
 
 A frequent use case for xDS involves client subscription to multiple resources from the xDS management server. Currently, if the xDS management server cannot provide a subset of these resources, the client experiences a complete loss of visibility. This can occur for various reasons, including but not limited to potential permission issues. This lack of granularity in the response can lead to significant operational challenges. 
 
-The xDS protocol does have a way for Envoy/gRPC to NACK responses back from the xDS Management server. The NACK response also contains an error_detail which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client.
-
-This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the applications to debug these issues without escalation. 
-
-
+The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the applications to debug these issues without escalation. 
 
 ### Related Proposals:
 
@@ -86,12 +81,11 @@ The xDS Management server is only expected to return the error message once rath
 
 It is possible to subscribe to all resources by a client using a wildcard or “” resource name.The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources. 
 
-For errors associated with wildcard(“” for legacy or “*”) and for xDS-TP glob collections; the control plane error_details resource name will match the relevant wildcard request(“” or “*” or xDS-TP glob collections). This can be used by the control plane to indicate an error with the collection as a whole.
-For errors associated with specific individual resources that match the glob,
-The resource name should be the specific resource name associated with the error 
-OR
-The control could just not use this mechanism for wildcard subscriptions, because if the client doesn't have permission to access a resource, then it probably shouldn't be considered to match the wildcard subscription to begin with.
-
+1. For errors associated with wildcard("" for legacy or "*") and for xDS-TP glob collections; the control plane `error_details` resource name will match the relevant wildcard request("" or "*" or xDS-TP glob collections). This can be used by the control plane to indicate an error with the collection as a whole.
+2. For errors associated with specific individual resources that match the glob,
+    - The resource name should be the specific resource name associated with the error 
+            OR
+    - The control could just not use this mechanism for wildcard subscriptions, because if the client doesn't have permission to access a resource, then it probably shouldn't be considered to match the wildcard subscription to begin with.
 
 ## Rationale
 
@@ -101,13 +95,13 @@ This section documents limitations and design alternatives that were considered.
 
 The major alternative to this proposal is to use Wrapped Resources. 
 
-xDS supports passing resources via a wrapped Resource Container. This is the default for the Incremental xDS protocol and is controlled via the client feature xds.config.supports-resource-in-sotw for SoTW. 
+xDS supports passing resources via a wrapped Resource Container. This is the default for the Incremental xDS protocol and is controlled via the client feature `xds.config.supports-resource-in-sotw` for SoTW. 
 
 In this proposal the error information is directly passed as part of the resource field in the Resource Container, using the artifact that the field currently is protobuf.Any. This enables us to designate the resource as either a `ResourceError` if the xDS management server encountered problems, or as the actual `Resource` if no errors occurred. 
 
 #### Backward Compatibility
 
-To avoid possible confusion with this behavior, it will be protected with a client feature similar to  supports-resource-in-sotw  called supports-resource-error-unwrapping. 
+To avoid possible confusion with this behavior, it will be protected with a client feature similar to  `supports-resource-in-sotw`  called `supports-resource-error-unwrapping`. 
 
 Note: This should also be documented here: https://www.envoyproxy.io/docs/envoy/latest/api/client_features#currently-defined-client-features
 
@@ -116,9 +110,8 @@ Note: This should also be documented here: https://www.envoyproxy.io/docs/envoy/
 
 This approach has two major drawbacks compared to the chosen approach:
 
-This alternative would not work for non-wrapped resources
-It introduces a backwards compatibility issue, which adding a new field wouldn’t have as clients would just ignore them. 
-
+* This alternative would not work for non-wrapped resources
+* It introduces a backwards compatibility issue, which adding a new field wouldn’t have as clients would just ignore them. 
 
 ## Implementation
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -11,13 +11,13 @@ This proposal introduces enhancements to the [xDS transport protocol](https://ww
 
 The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
 
-This proposal includes a new field for each subscribed Resource, called `ResourceError`. This field will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
+This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
 
 ## Background
 
-A frequent use case for xDS involves client subscription to multiple resources from the xDS management server. Currently, if the xDS management server cannot provide a subset of these resources, the client experiences a complete loss of visibility. This can occur for various reasons, including but not limited to potential permission issues. This lack of granularity in the response can pose significant operational challenges. 
+A frequent use case for xDS involves client subscription to multiple resources from the management server. Currently, if the xDS management server cannot provide a subset of these resources, the client experiences a complete loss of visibility. This can occur for various reasons, including but not limited to potential permission issues. This lack of granularity in the response can pose significant operational challenges for general client observability or debugging. 
 
-The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the applications to debug these issues without escalation. 
+The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the client applications to debug these issues without escalation. 
 
 ### Related Proposals:
 
@@ -90,7 +90,7 @@ The major alternative to this proposal is to use Wrapped Resources by using Reso
 
 ### Wrapped Resources
 
- xDS resource containers are the default protos used for the Incremental xDS protocol and it's usage in SoTW is controlled via the client feature `xds.config.supports-resource-in-sotw`. 
+xDS resource containers are the default protos used for the Incremental xDS protocol and it's usage in SoTW is controlled via the client feature `xds.config.supports-resource-in-sotw`. 
 
 In this proposal the error information is directly passed as part of the resource field in the Resource Container, using the artifact that the field is a protobuf.Any. This enables us to designate the resource as either a `ResourceError` if the xDS management server encountered problems, or as the actual `Resource` if no errors occurred. 
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -11,7 +11,7 @@ This proposal introduces a mechanism for the xDS control plane to communicate er
 
 The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
 
-This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
+This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client may use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
 
 ## Background
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -57,10 +57,7 @@ string version_info = 1;
 // notifications in case of unavailability, permissions errors 
 // without the client having to wait for the `config fetch` timeout.
 repeated ResourceError resource_errors = N;
-
-// NEW FIELD
-// For non-existing resources.
-repeated ResourceName removed_resources = N;
+...
 }
 
 message DeltaDiscoveryResponse {
@@ -72,14 +69,20 @@ message DeltaDiscoveryResponse {
 ```
 
 ### Protocol Behavior
-Clients that support this mechanism will use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application(i.e. fail a data plane request with a useful error message instead of a timeout). With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
+Clients that support this mechanism will use this additional field to obtain error information for resources that the xDS server couldn’t provide. When this message is received, the xDS client should cancel any resource timers for the indicated resource. The client should then handle the error based on the status code, as follows (using RFC-2119 terminology):
+
+The following codes will be interpretted by clients as transient failures, meaning that the client MUST continue to use previously cached versions of the resource if it has them: UNAVAILABLE, INTERNAL, UNKNOWN.
+
+The following codes will be interpretted by clients as data errors, which the client MAY handle by dropping any previously cached resource: NOT_FOUND, PERMISSION_DENIED
+The behavior for any other status code is undefined. Data planes SHOULD treat these as transient failures, but future xRFCs may impose additional semantics on them.
+Note that an error with status code NOT_FOUND will be interpreted to mean that the resource does not exist. The client should handle this case exactly the same way that it would if the does-not-exist timer fired or if it receives removed_resources or removed_resource_names in the incremental protocol variant.
 
 The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed 
 resource in every response. 
 
 As the clients don't look at this field right now, the xDS management server must assume the responses remain valid for older clients for backward compatibility. For example, in the SotW variant, LDS and CDS are required to send all subscribed resources in every response, which means that if a control plane wants to send an update for one of these resource types that indicates an error with a particular resource, it must still include all of the subscribed resources that it can return in that response.
 
-### Wildcard Resources and Glob Collections
+### Wildcard Subscriptions and Glob Collections
 
 For legacy wildcard subscriptions or [glob collections](TP1-xds-transport-next.md#glob), the control plane is responsible for deciding which resources are included in the subscription. The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources. 
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -21,7 +21,7 @@ The xDS protocol does have a way for xDS clients to NACK responses back from the
 
 ### Related Proposals:
 
-N/A
+* [TP1-xds-transport-next.md](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md)
 
 ## Proposal
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -7,7 +7,7 @@ TP3: xds-error-propagation
 
 ## Abstract
 
-This proposal introduces enhancements to the [xDS transport protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol), focusing on improved debugging and observability. Specifically, it outlines a mechanism for the control plane to communicate error information directly to xDS clients.
+This proposal introduces a mechanism for the xDS control plane to communicate errors to clients in cases where the control plane is unable to send a resource that the client has subscribed to (e.g., because the resource does not exist or because the client lacks permission for the resource), thus improving debuggability. In addition, this also provides a mechanism in the [SotW protocol variants](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#variants-of-the-xds-transport-protocol) for the control plane to explicitly tell the client that a resource does not exist, without the client needing to wait for a [15-second does-not-exist timeout](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#knowing-when-a-requested-resource-does-not-exist), similar to what already exists in the incremental protocol variants.
 
 The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
 
@@ -15,9 +15,10 @@ This proposal includes a new field for each subscribed Resource, called `Resourc
 
 ## Background
 
-A frequent use case for xDS involves client subscription to multiple resources from the management server. Currently, if the xDS management server cannot provide a subset of these resources, the client experiences a complete loss of visibility. This can occur for various reasons, including but not limited to potential permission issues. This lack of granularity in the response can pose significant operational challenges for general client observability or debugging. 
+In cases where the xDS control plane is unable to send a resource that the client has subscribed to (e.g., because the resource does not exist or because the client lacks permission for the resource), the control plane does not have a good mechanism for conveying the cause of the error to the client. The only existing mechanism for conveying a specific error message to a client is for the control plane to close the entire xDS stream with a non-OK status. However, this is often undesirable behavior, because a problem with a single resource will also stop the client from receiving updates from other, unrelated resources.
+Therefore, most control planes are forced to simply act as if the resource does not exist. In this situation, the client does not know why the control plane was unable to send the resource, so it cannot report useful information to human operators (e.g., via logs or request failure messages), which makes debugging challenging.
 
-The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the client applications to debug these issues without escalation. 
+Note that while the incremental protocol variants can explicitly indicate to the client that a subscribed resource does not exist, the SotW protocol variants require the client to use a 15-second does-not-exist timeout. This adds an unnecessary delay before the client can react to the error. It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.
 
 ### Related Proposals:
 
@@ -73,7 +74,7 @@ message DeltaDiscoveryResponse {
 ### Protocol Behavior
 The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application. With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
 
-The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. 
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed resource in every response. 
 
 ### Wildcard Resources
 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -3,7 +3,7 @@ TP3: xds-error-propagation
 * Author(s): anicr7
 * Approvers: markdroth, adisuissa
 * Implemented in: <xDS client, ...>
-* Last updated: 2024-10-25
+* Last updated: 2024-12-12
 
 ## Abstract
 
@@ -26,7 +26,7 @@ Note that while the incremental protocol variants can explicitly indicate to the
 
 ## Proposal
 
-The proposal introduces two new fields to the DiscoveryResponse proto. The first, `resource_errors`, will detail Resource-specific errors and be included in both SotW and Incremental xDS protocols. The second, `removed_resources`, will specifically identify resources that no longer exist, addressing a current gap in the SotW protocol's signaling capabilities. 
+The proposal introduces a new field to the DiscoveryResponse/DeltaDiscoveryResponse proto. This new field `resource_errors` will detail Resource-specific errors and be included in both SotW and Incremental xDS protocols. 
 
 ```textproto
 // New Proto Message
@@ -68,6 +68,23 @@ message DeltaDiscoveryResponse {
 }
 ```
 
+A corresponding new field will also be added to CSDS which should be treated the same as `CLIENT_NACKED` but also indicates a server error received via `resource_errors`.
+
+```
+service ClientStatusDiscoveryService {
+
+  // Config status from a client-side view.
+  enum ClientConfigStatus {
+    ...
+
+    // NEW FIELD
+    // Should be treated basically the same as `CLIENT_NACKED`.
+    // Provides the previously cached version of the resource but also indicates a server error that was recieved using resource_errors field.
+    CLIENT_RECEIVED_ERROR = 4;
+  }
+}
+```
+
 ### Protocol Behavior
 Clients that support this mechanism will use this additional field to obtain error information for resources that the xDS server couldn’t provide. When this message is received, the xDS client should cancel any resource timers for the indicated resource. The client should then handle the error based on the status code, as follows (using [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) terminology):
 
@@ -75,7 +92,7 @@ Clients that support this mechanism will use this additional field to obtain err
   * The following codes will be interpretted by clients as data errors, which the client MAY handle by dropping any previously cached resource: NOT_FOUND, PERMISSION_DENIED
   * The behavior for any other status code is undefined. Data planes SHOULD treat these as transient failures, but future xRFCs may impose additional semantics on them.
 
-Note that an error with status code NOT_FOUND will be interpreted to mean that the resource does not exist. The client should handle this case exactly the same way that it would if the does-not-exist timer fired or if it receives removed_resources or removed_resource_names in the incremental protocol variant.
+Note that an error with status code NOT_FOUND will be interpreted to mean that the resource does not exist. The client should handle this case exactly the same way that it would if the does-not-exist timer fired.
 
 The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed 
 resource in every response. The set of resources with errors and the set of valid resources must not intersect. 
--- a/proposals/TP3-xds-error-propagation.md
+++ b/proposals/TP3-xds-error-propagation.md
@@ -1,6 +1,6 @@
 TP3: xds-error-propagation
 ----
-* Author(s): Anirudh Ramachandra
+* Author(s): anicr7
 * Approvers: htuch, markdroth, adisuissa
 * Implemented in: <xDS client, ...>
 * Last updated: 2024-10-25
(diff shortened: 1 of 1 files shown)


proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+// can provide to notify the client of issues for the resources that 
+// are subscribed.
+// This allows the xDS management server to provide optional 
+// notifications in case of unavailability, permissions errors 
+// without the client having to wait for the `config fetch` timeout.
+repeated ResourceError resource_errors = N;
+
+// NEW FIELD
+// For non-existing resources.
+repeated ResourceName removed_resources = N;
```
howardjohn, line 62: Isn't this implied by the lack of existence in `resources`?

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+
+  // NEW FIELD
+  repeated ResourceError resource_errors = N;
+}
+```
+
+### Protocol Behavior
+The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application. With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
+
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned.
```
howardjohn, line 76: Only _expected_ or is _required not to_?

proposals/TP3-xds-error-propagation.md
anicr7: Currently if the xDS management server doesn't provide any resources its not clear if the resources are not present or if there was just a delay in sending these resources across to the client. The client currently can't differentiate between the two scenarios. 

By explicitly providing a list of resources that don't exist, the client can stop the timer and conclude that the resources are missing.

proposals/TP3-xds-error-propagation.md
anicr7: This statement is to clarify that the xDS management server is free to send the error back for future messages too rather than just for the first one. I am not sure if we want to codify this with a stronger statement  i.e. the xDS management server should return the error message *only once*. 

Do you see a benefit by making this a requirement?

proposals/TP3-xds-error-propagation.md
howardjohn: I mean I kind of get it.. but also this has nothing to do with error propagation really? It feels like it's slipping in "retrofit sotw to have some delta functionality" into an error propagation proposal when they are really orthogonal

anicr7: The major reason for this is to `propagate the error` that the resource is technically missing. If we don't cover this scenario we still have this gap for SotW. 

I do resonate with your point that this might be overstepping the scope of this error propagation proposal. I am ok adding this into another specific xRFC.

@markdroth in case he has additional thoughts.

proposals/TP3-xds-error-propagation.md
howardjohn: No I think it's good to not make it "exactly once", just wanted to clarify. thanks

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
+
+This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
+
+## Background
+
+A frequent use case for xDS involves client subscription to multiple resources from the management server. Currently, if the xDS management server cannot provide a subset of these resources, the client experiences a complete loss of visibility. This can occur for various reasons, including but not limited to potential permission issues. This lack of granularity in the response can pose significant operational challenges for general client observability or debugging.
```
markdroth, line 18: Suggested wording for this section:

"""
In cases where the xDS control plane is unable to send a resource that the client has subscribed to (e.g., because the resource does not exist or because the client lacks permission for the resource), the control plane does not have a good mechanism for conveying the cause of the error to the client.

The only existing mechanism for conveying a specific error message to a client is for the control plane to close the entire xDS stream with a non-OK status.  However, this is often undesirable behavior, because a problem with a single resource will also stop the client from receiving updates from other, unrelated resources.

Therefore, most control planes are forced simply act as if the resource does not exist.  In this situation, the client does not know why the control plane was unable to send the resource, so it cannot report useful information to human operators (e.g., via logs or request failure messages), which makes debugging challenging.

Note that while the incremental protocol variants can explicitly indicate to the client that a subscribed resource does not exist, the SotW protocol variants require the client to use a [15-second does-not-exist timeout](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol#knowing-when-a-requested-resource-does-not-exist).  This adds an unnecessary delay before the client can react to the error.  It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.
"""

proposals/TP3-xds-error-propagation.md
markdroth: I can see how this might seem tangential, but I think it's an important part of this proposal.  The overall goal here is for the control plane to be able to explicitly indicate errors to the client, so that the client can tell the difference between various types of errors -- e.g., I think we want to understand the difference between "does not exist" and "permission denied" and (for SotW) "the control plane is just being slow".

That having been said, in our internal discussions, we did go back and forth a few times as to whether we needed a separate field for does-not-exist or whether we should just encode an error with status NOT_FOUND in that case.  I think the core question here is whether we'd want the client to react in a different way to does-not-exist than to an error like permission-denied.  Today, gRPC does have special handling for does-not-exist: it causes us to deconfigure the channel instead of just sticking with the previously cached resource, if any, which we would on other types of errors (e.g., a connectivity error with the control plane or receiving an invalid resource).  But it's worth noting that today, the control plane would need to treat permission-denied as does-not-exist, so the client would use that same special behavior in this case.  I guess we should explicitly decide what behavior we want for cases like permission-denied.

So I see a few options here:

1. gRPC treats any error from the control plane the same as does-not-exist, just with a better error message.  In this case, we don't really need the separate `removed_resources` field, because we can encode both cases in the `resource_errors` field.
2. gRPC treats does-not-exist differently than other errors.  We have two ways of reflecting this in the transport protocol:
   a. We can encode does-not-exist via this separate `removed_resources` field, as this xRFC currently indicates.
   b. We can not add the `removed_resources` field and instead rely on NOT_FOUND status in `resource_errors`.  This seems a little sub-optimal, because it imbues one particular status code with special meaning, which is a lot less clear.

@ejona86 and @dfawley, thoughts...?

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+message DeltaDiscoveryResponse {
+  …
+
+  // NEW FIELD
+  repeated ResourceError resource_errors = N;
+}
+```
+
+### Protocol Behavior
+The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application. With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired.
```
markdroth, line 74: Given that all currently existing clients don't know anything about this field, I don't think we can say that clients *must* use the field.  I think we need to make the backward-compatibility story clear here.  I think the approach here will be that control planes are free to send this field, but they need to understand that while newer clients will use the field, older clients will not use it.

This means that any responses sent by the control plane must be valid even for older clients.  For example, in the SotW variant, LDS and CDS are required to send all subscribed resources in every response, which means that if a control plane wants to send an update for one of these resource types that indicates an error with a particular resource, it must still include all of the subscribed resources that it *can* return in that response.

markdroth, line 74: I think we should probably expand a bit on the "convey the error message to the application" part.  I think that's mainly oriented toward failing data plane requests with a useful error message, and we should be clear that that's what we're referring to.

Also, the question of whether we always do this depends on the question above regarding whether we actually want clients to handle errors like permission-denied the same as does-not-exist.

proposals/TP3-xds-error-propagation.md
markdroth: I think that if we want to give the control plane the option of not resending every time, then we need to require the client to remember the state.

We should also explicitly state that even for LDS and CDS, where the control plane is required to send every subscribed resource in every response, the control plane does not need to resend the error messages in every response.  The client is required to remember the most recent error state for each resource even for LDS and CDS.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. 
+
+### Wildcard Resources
+
+It is possible to subscribe to all resources by a client using a wildcard or "" resource name. The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources. 
+
+1. For errors associated with wildcard("" for legacy or "*") and for xDS-TP glob collections; the control plane `error_details` resource name will match the relevant wildcard request("" or "*" or xDS-TP glob collections). This can be used by the control plane to indicate an error with the collection as a whole.
```
markdroth, line 82: The asterisks here are being interpretted as markdown formatting rather than a literal.  Suggest using backquotes instead of double quotes to avoid this.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the client applications to debug these issues without escalation. 
+
+### Related Proposals:
```
markdroth, line 22: This should list xRFC TP1, since we're referring to glob collections below.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+### Protocol Behavior
+The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application. With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
+
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. 
+
+### Wildcard Resources
+
+It is possible to subscribe to all resources by a client using a wildcard or "" resource name. The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources.
```
markdroth, line 80: This first sentence should also include glob collections, with a link to https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md#glob.

proposals/TP3-xds-error-propagation.md
markdroth, line 82: My inclination is to say that we support this case only for glob collections, not for old-style wildcard subscriptions.  We want people to migrate away from old-style wildcard subscriptions anyway.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+* This alternative would not work for non-wrapped resources
+* It introduces a backwards compatibility issue, which adding a new field wouldn’t have as clients would just ignore them. 
+
+## Implementation
+
+This will probably be implemented in gRPC before Envoy.
+
+## Open issues (if applicable)
+
+### Possible improvements to NACKs
```
markdroth, line 116: I think this issue has nothing to do with this proposal, since it's talking about a very different problem.  I suggest just removing this.

markdroth (commented):
Thanks for writing this up, @anicr7!

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
+TP3: xds-error-propagation
+----
+* Author(s): anicr7
+* Approvers: htuch, markdroth, adisuissa
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+* Implemented in: <xDS client, ...>
+* Last updated: 2024-10-25
+
+## Abstract
+
+This proposal introduces enhancements to the [xDS transport protocol](https://www.envoyproxy.io/docs/envoy/latest/api-docs/xds_protocol), focusing on improved debugging and observability. Specifically, it outlines a mechanism for the control plane to communicate error information directly to xDS clients.
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,122 @@
...
+
+The xDS protocol does have a way for xDS clients to NACK responses back from the xDS Management server. The NACK response also contains an `error_detail` field which the Management Server can use to extract further information about the rejection. But this NACK’ing mechanism is restricted to the client i.e. there is no way for Management Server to actually convey a notification(Ex: unavailability, permission issues etc) regarding some or all the resource requests that are being subscribed by the client. This eventually leads to the client timeouts which, although it conveys an error back to the application, it misses the actual context for the issue. In most cases the xDS Management Servers might not even be accessible for the client applications to debug these issues without escalation.
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: I don't have a strong opinion about supporting old-style wildcard, @adisuissa any preferences?

proposals/TP3-xds-error-propagation.md
anicr7: Done, also added the example that you mentioned into this section for clarification.

anicr7: Done.

proposals/TP3-xds-error-propagation.md
```diff
@@ -71,15 +72,18 @@ message DeltaDiscoveryResponse {
...
 
 ### Wildcard Resources
 
-It is possible to subscribe to all resources by a client using a wildcard or "" resource name. The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources. 
+It is possible to subscribe to all resources by a client using a wildcard or "" resource name or a [global collection](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md). The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources.
```
markdroth, line 84: Suggest replacing this first sentence with the following:

"""
For legacy wildcard subscriptions or [glob collections](TP1-xds-transport-next.md#glob), the control plane is responsible for deciding which resources are included in the subscription.
"""

Note that the link should point to the specific section in TP1 about glob collections: `TP1-xds-transport-next.md#glob`.

proposals/TP3-xds-error-propagation.md
```diff
@@ -71,15 +72,18 @@ message DeltaDiscoveryResponse {
...
 
-The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. 
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed 
+resource in every response. 
+
+As the clients don't look at this field right now, the xDS management server must assume the responses remain valid for older clients for backward compatibility. For example, in the SotW variant, LDS and CDS are required to send all subscribed resources in every response, which means that if a control plane wants to send an update for one of these resource types that indicates an error with a particular resource, it must still include all of the subscribed resources that it can return in that response.
 
 ### Wildcard Resources
```
markdroth, line 82: "Wildcard Subscriptions and Glob Collections"

markdroth (commented):
This looks good overall!  I have a few more cosmetic comments, but the two main open questions are the ones from my last review pass on which I want to get some additional input.

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -71,15 +72,18 @@ message DeltaDiscoveryResponse {
 ```
 
 ### Protocol Behavior
-The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application. With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
+All client are eventually expected to use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application(i.e. fail a data plane request with a useful error message instead of a timeout). With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired.
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -1,27 +1,28 @@
...
+Note that while the incremental protocol variants can explicitly indicate to the client that a subscribed resource does not exist, the SotW protocol variants require the client to use a 15-second does-not-exist timeout. This adds an unnecessary delay before the client can react to the error. It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.
 
 ### Related Proposals:
 
-N/A
+* [TP1-xds-transport-next.md](https://github.com/cncf/xds/blob/main/proposals/TP1-xds-transport-next.md)
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,118 @@
...
+message DeltaDiscoveryResponse {
+  …
+
+  // NEW FIELD
+  repeated ResourceError resource_errors = N;
+}
+```
+
+### Protocol Behavior
+Clients that support this mechanism will use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application(i.e. fail a data plane request with a useful error message instead of a timeout). With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired.
```
markdroth, line 75: As per my other comment, I suggest the following wording for this paragraph:

"""
Clients that support this mechanism will use this additional field to obtain error information for resources that the xDS server couldn’t provide. When this message is received, the xDS client should cancel any resource timers for the indicated resource.  The client should then handle the error based on the status code, as follows (using [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) terminology):
- The following codes will be interpretted by clients as transient failures, meaning that the client MUST continue to use previously cached versions of the resource if it has them: `UNAVAILABLE`, `INTERNAL`, `UNKNOWN`
- The following codes will be interpretted by clients as data errors, which the client MAY handle by dropping any previously cached resource: `NOT_FOUND`, `PERMISSION_DENIED`
- The behavior for any other status code is undefined.  Data planes SHOULD treat these as transient failures, but future xRFCs may impose additional semantics on them.

Note that an error with status code `NOT_FOUND` will be interpreted to mean that the resource does not exist.  The client should handle this case exactly the same way that it would if the does-not-exist timer fired or if it receives `removed_resources` or `removed_resource_names` in the incremental protocol variant.
'""

anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -72,16 +72,16 @@ message DeltaDiscoveryResponse {
...
 
 The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed 
 resource in every response. 
 
 As the clients don't look at this field right now, the xDS management server must assume the responses remain valid for older clients for backward compatibility. For example, in the SotW variant, LDS and CDS are required to send all subscribed resources in every response, which means that if a control plane wants to send an update for one of these resource types that indicates an error with a particular resource, it must still include all of the subscribed resources that it can return in that response.
 
-### Wildcard Resources
+### Wildcard Resources and Glob Collections
```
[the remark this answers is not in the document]
anicr7: Done

proposals/TP3-xds-error-propagation.md
anicr7: Done, changed

proposals/TP3-xds-error-propagation.md
```diff
@@ -72,14 +69,19 @@ message DeltaDiscoveryResponse {
 ```
 
 ### Protocol Behavior
-Clients that support this mechanism will use this additional field to obtain notification for resources the xDS Management server couldn’t procure. The xDS client should cancel any resource timers once this message is received and convey the error message to the application(i.e. fail a data plane request with a useful error message instead of a timeout). With the addition of this field, when a client receives an explicit error or does-not-exist indicator from the management server, it should react the same way it would have if its does-not-exist timer fired. 
+Clients that support this mechanism will use this additional field to obtain error information for resources that the xDS server couldn’t provide. When this message is received, the xDS client should cancel any resource timers for the indicated resource. The client should then handle the error based on the status code, as follows (using RFC-2119 terminology):
```
markdroth, line 72: Please make "RFC-2119" a link to https://datatracker.ietf.org/doc/html/rfc2119.

proposals/TP3-xds-error-propagation.md
```diff
@@ -72,14 +69,19 @@ message DeltaDiscoveryResponse {
...
+
+  * The following codes will be interpretted by clients as transient failures, meaning that the client MUST continue to use previously cached versions of the resource if it has them: UNAVAILABLE, INTERNAL, UNKNOWN.
+  * The following codes will be interpretted by clients as data errors, which the client MAY handle by dropping any previously cached resource: NOT_FOUND, PERMISSION_DENIED
+  * The behavior for any other status code is undefined. Data planes SHOULD treat these as transient failures, but future xRFCs may impose additional semantics on them.
+Note that an error with status code NOT_FOUND will be interpreted to mean that the resource does not exist. The client should handle this case exactly the same way that it would if the does-not-exist timer fired or if it receives removed_resources or removed_resource_names in the incremental protocol variant.
```
markdroth, line 77: Please add a newline before this, so that it shows up as a separate paragraph and not as part of the preceding bullet.

proposals/TP3-xds-error-propagation.md
markdroth: @adisuissa, any concern about not supporting old-style wildcard subscriptions here?

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+### Wildcard Subscriptions and Glob Collections
+
+For legacy wildcard subscriptions or [glob collections](TP1-xds-transport-next.md#glob), the control plane is responsible for deciding which resources are included in the subscription. The control plane in this case can provide error details for two different use cases. One when the issue is with the glob itself or later on when the issue is specific to individual resources. 
+
+1. For errors associated with wildcard("" for legacy or `*`) and for xDS-TP glob collections; the control plane `error_details` resource name will match the relevant wildcard request("" or `*` or xDS-TP glob collections). This can be used by the control plane to indicate an error with the collection as a whole.
+2. For errors associated with specific individual resources that match the glob,
+    * The resource name should be the specific resource name associated with the error  OR
+    * The control could just not use this mechanism for wildcard subscriptions, because if the client doesn't have permission to access a resource, then it probably shouldn't be considered to match the wildcard subscription to begin with.
+
+## Rationale
```
adisuissa, line 94:
```suggestion
## Alternatives
```

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+To avoid possible confusion with this behavior, it must be protected with a client feature similar to `supports-resource-in-sotw` called `supports-resource-error-unwrapping`. 
+
+Note: This should also be documented here: https://www.envoyproxy.io/docs/envoy/latest/api/client_features#currently-defined-client-features
+
+#### Why not this approach?
+
+This approach has two major drawbacks compared to the chosen approach:
+
+* This alternative would not work for non-wrapped resources
+* It introduces a backwards compatibility issue, which adding a new field wouldn’t have as clients would just ignore them.
```
adisuissa, line 115: Why is that? Isn't this the same as adding the newly proposed field?

proposals/TP3-xds-error-propagation.md
adisuissa: While Envoy is configured to use "*" for wildcard, it doesn't, as this may break current servers. It may have been possible if there was a plan to upgrade the protocol to v4.
Specifically for this field, it would be possible to have it asymmetric, that is Envoy subscribes using the old wildcard (""), and receives a response where the resource_name in the resource_errors is the new wildcard ("*").

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+  // NEW FIELD
+  repeated ResourceError resource_errors = N;
+}
+```
+
+### Protocol Behavior
+Clients that support this mechanism will use this additional field to obtain error information for resources that the xDS server couldn’t provide. When this message is received, the xDS client should cancel any resource timers for the indicated resource. The client should then handle the error based on the status code, as follows (using [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) terminology):
+
+  * The following codes will be interpretted by clients as transient failures, meaning that the client MUST continue to use previously cached versions of the resource if it has them: UNAVAILABLE, INTERNAL, UNKNOWN.
```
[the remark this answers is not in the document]
markdroth: I can think of several reasons why the client might later see an error after having previously seen a valid resource.  For example, maybe the ADS stream gets closed and the client winds up reconnecting to a different xDS server instance that has a different view of the resources.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+  * The following codes will be interpretted by clients as data errors, which the client MAY handle by dropping any previously cached resource: NOT_FOUND, PERMISSION_DENIED
```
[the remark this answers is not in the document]
markdroth: This is not a behavior change.

Note that use of "MAY" here, which is intended in the [RFC-2119](https://datatracker.ietf.org/doc/html/rfc2119) sense.  Envoy doesn't have to do this if you don't want to.

gRPC has an option today where it can drop previously cached resources when it sees a resource deletion.  This option already exists, so this isn't introducing anything new.

I don't see any conflict between this and the eventual consistency model.  If the resource comes into existence again later, the client will start using it.  All we're really doing here is not treating "does not exist" any differently than any other value of the resource.

proposals/TP3-xds-error-propagation.md
markdroth: This is basically saying that it requires the client capability.  Without that, the server cannot send the new fields to the client, because older clients would not know how to interpret a `google.rpc.Status` message inside the `Any` field where the resource should be.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Rationale
+
+The major alternative to this proposal is to use Wrapped Resources by using Resource Containers defined in https://www.envoyproxy.io/docs/envoy/latest/xds/core/v3/resource.proto#xds-core-v3-resource. 
+
+### Wrapped Resources
+
+xDS resource containers are the default protos used for the Incremental xDS protocol and it's usage in SoTW is controlled via the client feature `xds.config.supports-resource-in-sotw`. 
+
+In this proposal the error information is directly passed as part of the resource field in the Resource Container, using the artifact that the field is a protobuf.Any. This enables us to designate the resource as either a `ResourceError` if the xDS management server encountered problems, or as the actual `Resource` if no errors occurred.
```
markdroth, line 102: I think this alternative would have encoded it as a `google.rpc.Status`, not a `ResourceError`.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+Note that an error with status code NOT_FOUND will be interpreted to mean that the resource does not exist. The client should handle this case exactly the same way that it would if the does-not-exist timer fired or if it receives removed_resources or removed_resource_names in the incremental protocol variant.
+
+The xDS Management server is only expected to return the error message once rather than throughout for future responses. The client is expected to remember the error message until either a new error message is returned or the resource is returned. This includes LDS and CDS where the control plane is required to send every subscribed 
+resource in every response. 
+
+As the clients don't look at this field right now, the xDS management server must assume the responses remain valid for older clients for backward compatibility. For example, in the SotW variant, LDS and CDS are required to send all subscribed resources in every response, which means that if a control plane wants to send an update for one of these resource types that indicates an error with a particular resource, it must still include all of the subscribed resources that it can return in that response.
```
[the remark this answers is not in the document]
markdroth: I'm fine with clarifying that the set of resources with errors and the set of resources with contents must not intersect.  I think this is already true for the `resources_removed` field in `DeltaDiscoveryResponse`.

I think in almost all cases, the errors we're sending back here will wind up being smaller than the corresponding resources would have been.  If size wasn't a concern for the resources themselves, it shouldn't be a concern here either.  And ultimately, the control plane can simply choose not to send errors for each individual resource here, as described in the following section.

I think that if there's a case where returning the resource names would be a problem, the server can simply choose to send an error for the collection as a whole, as described in the following section.  I don't know of a concrete case where there would be a security problem for only a strict subset of resources where it would be important not to send the resource names.

proposals/TP3-xds-error-propagation.md
markdroth: The intent for introducing `*` was that Envoy would use `*` for wildcard if it was *also* subscribing to on-demand resources.  Envoy was not going to do that if it was *not* subscribing to on-demand resources, since that preserved backward-compatibility for the common case with existing control plane implementations.  See discussion in https://github.com/envoyproxy/envoy/pull/16861.

I agree that we could support wildcard subscriptions here, and always using `*` here would be one way to do that.  My question here is really about whether that's really necessary in the first place.  It seems simpler to just say that this mechanism doesn't support wildcard subscriptions, and if you want to use this, you need to upgrade to xdstp collections (either list or glob).  Is there any reason that won't work?  Do we actually have any cases where we'd need this with wildcard subscriptions?

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
markdroth: BTW, I should clarify that this doesn't affect the cache at the xDS transport protocol layer; it's really referring to state kept at the data model layer.  The watcher still needs to cache the previous resource in either case, but an individual watcher (e.g., in a Cluster implementation) may decide to that it doesn't want to continue to use that cached data if it sees a particular error.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
howardjohn: As a control plane implementation, I don't think I could responsibly decide to use an error code if it the behavior of it is undefined. I would instead use mechanisms that are clearly defined (which already mostly exists). It doesn't seem like we would want this ambiguity?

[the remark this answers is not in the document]
markdroth: I think individual data planes are going to handle certain error cases differently than others, as evidenced by the fact that gRPC can be configured to ignore resource deletions and Envoy cannot.  I think all we can do here is document what options data planes are allowed so that control planes can make an informed decision about what codes to send.  If you want to guarantee that the client will always keep using the previously cached version, if any, then there are codes you can use that will guarantee that behavior.

With respect to leaving certain codes undefined, the reasoning was basically that we can't think of any reasons why any of the other codes would actually be used for per-resource errors, and it didn't seem wise to impose semantics for those codes in the absence of concrete use cases.  If a control plane wants to be safe, it can just avoid those codes entirely, and then the issue doesn't arise.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
markdroth: This made me realize that we also need to define the CSDS behavior here.  That doesn't affect Envoy, since Envoy doesn't support CSDS yet, but gRPC does.

I think we need to add a new value to the [`ClientConfigStatus` enum](https://github.com/envoyproxy/envoy/blob/22e7328305b8fc682644e1bfbbc33f93b856993d/api/envoy/service/status/v3/csds.proto#L60), let's call it `CLIENT_RECEIVED_ERROR`.  This should be treated basically the same as `CLIENT_NACKED`: it will provide the previously cached version of the resource but also indicate the server error that we got in the [`error_state` field](https://github.com/envoyproxy/envoy/blob/22e7328305b8fc682644e1bfbbc33f93b856993d/api/envoy/service/status/v3/csds.proto#L170).

proposals/TP3-xds-error-propagation.md
adisuissa: I thought it would be in the [resource-wrapper](https://github.com/envoyproxy/envoy/blob/bb35c0c704d63109ebc15c29d96b3ae6f224b290/api/envoy/service/discovery/v3/discovery.proto#L336), and no resource will actually be sent. But I can see why there could be an issue with old-clients that assume that the `resource` (Any) field will be populated by some resource contents.

I think that another drawback of the alternative is for glob-collections, but not sure if it is worth mentioning.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
adisuissa: My personal take on having these error codes open to interpretation (w.r.t client's behavior) is that it may cause more harm than good. The motivation behind a protocol is to present a well-defined contract, and because the error-code value impact the xDS-related behavior, then they should be properly defined.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
adisuissa: > I can think of several reasons why the client might later see an error after having previously seen a valid resource. For example, maybe the ADS stream gets closed and the client winds up reconnecting to a different xDS server instance that has a different view of the resources.

The issue is that the error may be intermittent, because the newly connected server isn't aware of the resource.



> BTW, I should clarify that this doesn't affect the cache at the xDS transport protocol layer; it's really referring to state kept at the data model layer. The ~watcher~ transport protocol layer still needs to cache the previous resource in either case, but an individual watcher (e.g., in a Cluster implementation) may decide to that it doesn't want to continue to use that cached data if it sees a particular error.

IMHO if the decision of whether a resource should or should not be used is left to the resource owner that would be a violation of the protocol and may lead to inconsistencies between different data-plane implementations.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
+
+This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
+
+## Background
+
+In cases where the xDS control plane is unable to send a resource that the client has subscribed to (e.g., because the resource does not exist or because the client lacks permission for the resource), the control plane does not have a good mechanism for conveying the cause of the error to the client. The only existing mechanism for conveying a specific error message to a client is for the control plane to close the entire xDS stream with a non-OK status. However, this is often undesirable behavior, because a problem with a single resource will also stop the client from receiving updates from other, unrelated resources.
```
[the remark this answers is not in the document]
adisuissa: > I think the server always knows whether the resource exists at a given point in time. It may be that it will soon discover that the resource has come into existance, but that's a change that it should react to when it occurs. Prior to that point, the resource doesn't exist, and the server should react accordingly.

I think that in eventually consistent distributed systems, in cases where the server subscribes to get the resource from a different source-of-truth, then the server may not know if the resource exists or not until some timeout occurs. If the server is sending "resource doesn't exists" too early, then the client will update the resource state accordingly.

> I don't think anything in this proposal actually affects the fallback decision. This proposal provides a better mechanism to determine that a resource doesn't exist, but regardless of what mechanism we use to determine that, an individual resource not existing does not affect the fallback decision. We don't ever want to fall back just because a single resource doesn't exist.

I'm missing to see where a single resource doesn't exist is discussed here, the above example also applies to cases where all resources of some type are temporarily unknown by the (cold) server, and it sends "unknown" for all the resources before it actually learns about them.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+Therefore, most control planes are forced to simply act as if the resource does not exist. In this situation, the client does not know why the control plane was unable to send the resource, so it cannot report useful information to human operators (e.g., via logs or request failure messages), which makes debugging challenging.
+
+Note that while the incremental protocol variants can explicitly indicate to the client that a subscribed resource does not exist, the SotW protocol variants require the client to use a 15-second does-not-exist timeout. This adds an unnecessary delay before the client can react to the error. It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.
```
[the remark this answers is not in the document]
adisuissa: My main argument is regarding this:
> This adds an unnecessary delay before the client can react to the error. It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.

It is unclear to me how in an eventually consistent distributed system replying that something is "unknown" makes the system react correctly.
Specifically, if the server sends "unknown" before it actually learns about the existence resource, it will cause more issues. For example: when starting up, a client may wait until it receives the initially subscribed resources, before the data-plane can be used ("ready-mode"). If the server sends "unknown" too early, then the client will shift to the "ready-mode" state before it can actually serve the data-plane requests.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
markdroth: I understand this concern, but I don't think we can avoid it.  There is already a pre-existing mechanism in gRPC to allow different handling for the does-not-exist case (which today is also used to indicate permission problems), so there are already differences between data planes here.  All we're really doing in this proposal is providing a new way to indicate these cases, and documenting the existing behavior that data planes may be configured to provide.

For some discussion of the reasoning behind gRPC's behavior, see [gRFC A53](https://github.com/grpc/proposal/blob/master/A53-xds-ignore-resource-deletion.md) and my current draft of [gRFC A88](https://github.com/grpc/proposal/pull/466), which describes the changes we're going to make in gRPC to support this xRFC.

I'm open to hearing alternatives here, as long as they don't break existing gRPC functionality.

proposals/TP3-xds-error-propagation.md
markdroth: I think we'd have the same options for glob collections with this alternative that we do with the proposed approach.

proposals/TP3-xds-error-propagation.md
markdroth: @adisuissa, still need your input on this question about wildcard subscriptions.  (You might have missed it on your last review pass, since this thread is on an earlier commit, and the github UI is terrible...)

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+The objective of this proposal is to suggest a way for clients to receive feedback from xDS Management Servers in case of partial/drastic failures without closing any streams or connections.
+
+This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client must use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application.
```
adisuissa, line 14:
```suggestion
This proposal includes a new field for each subscribed Resource, called `ResourceError` which will provide detailed information for resource specific issues. The client may use this additional field to obtain notification for resources the xDS Management server couldn’t procure and provide necessary notification to the application. 
```

adisuissa (commented):
I still think that having the knowledge of the state of a resource (even if it exists) may be challenging in a distributed system that has a non-strong consistency model.
However, I understand that some systems may benefit from this field, and therefore it should be allowed but optional.

I'm resolving the comments.

proposals/TP3-xds-error-propagation.md
adisuissa: As discussed offline, it may be better to have this intentionally saying that the mechanism doesn't support wildcard subscriptions at the moment (and in the future this may be modified).

anicr7: Done, change the contents here to only refer to glob collections.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
anicr7: Done.

proposals/TP3-xds-error-propagation.md
anicr7: Fixed.

proposals/TP3-xds-error-propagation.md
anicr7: Done.

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,121 @@
...
+
+Note that while the incremental protocol variants can explicitly indicate to the client that a subscribed resource does not exist, the SotW protocol variants require the client to use a 15-second does-not-exist timeout. This adds an unnecessary delay before the client can react to the error. It has also led to cases where control plane slowness has caused clients to incorrectly react as if a resource does not exist.
+
+### Related Proposals:
+
+* [TP1-xds-transport-next.md](TP1-xds-transport-next.md)
+
+## Proposal
+
+The proposal introduces two new fields to the DiscoveryResponse proto. The first, `resource_errors`, will detail Resource-specific errors and be included in both SotW and Incremental xDS protocols. The second, `removed_resources`, will specifically identify resources that no longer exist, addressing a current gap in the SotW protocol's signaling capabilities.
```
markdroth, line 29: This paragraph still refers to `removed_resources`, which is no longer part of this proposal.

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
markdroth: @anicr7, I think we still need to include the CSDS change that I suggested in a comment earlier in this thread:

https://github.com/cncf/xds/pull/107#discussion_r1853159726

markdroth (commented):
This looks really good!  Just a couple of minor things left.

proposals/TP3-xds-error-propagation.md
anicr7: Done

proposals/TP3-xds-error-propagation.md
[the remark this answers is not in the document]
anicr7: Done

adisuissa (commented):
Please fix DCO

proposals/TP3-xds-error-propagation.md
```diff
@@ -0,0 +1,138 @@
...
+
+```
+service ClientStatusDiscoveryService {
+
+  // Config status from a client-side view.
+  enum ClientConfigStatus {
+    ...
+
+    // NEW FIELD
+    // Should be treated basically the same as `CLIENT_NACKED`.
```
markdroth, line 81: This comment should read as follows:

```
  // Client received an error from the control plane. The attached config
  // dump is the most recent accepted one. If no config is accepted yet,
  // the attached config dump will be empty.
```

markdroth (commented):
Just one small nit left!

For future reference, please don't force-push to a PR once a review has started, since that makes it hard for the review to see what's changed since their last review pass.  Thanks!

proposals/TP3-xds-error-propagation.md
anicr7: Done.

markdroth (approved):
This looks great!

Please fix CI, so that we can get this merged.  Thanks!

anicr7:
> This looks great!
> 
> Please fix CI, so that we can get this merged. Thanks!

@markdroth This should now be fixed!

PR #109, opened 2024-12-04
orca: update the orca validation annotation

wbpcode:


xds/data/orca/v3/orca_load_report.proto
```diff
@@ -21,7 +21,7 @@ message OrcaLoadReport {
 
   // Memory utilization expressed as a fraction of available memory
   // resources. This should be derived from the latest sample or measurement.
-  double mem_utilization = 2 [(validate.rules).double.gte = 0, (validate.rules).double.lte = 1];
+  double mem_utilization = 2 [(validate.rules).double.gte = 0];
```
wbpcode, line 24: The value may be larger than 1.0 when the usage exceeds the reporter dependent notion of soft limits, like the `application_utilization` did.

xds/data/orca/v3/orca_load_report.proto
```diff
@@ -30,12 +30,11 @@ message OrcaLoadReport {
...
   // Application specific requests costs. Each value is an absolute cost (e.g. 3487 bytes of
   // storage) associated with the request.
-  map<string, double> request_cost = 4;
+  map<string, double> request_cost = 4 [(validate.rules).map.values.double.gte = 0];
 
   // Resource utilization values. Each value is expressed as a fraction of total resources
   // available, derived from the latest sample or measurement.
-  map<string, double> utilization = 5
-      [(validate.rules).map.values.double.gte = 0, (validate.rules).map.values.double.lte = 1];
+  map<string, double> utilization = 5 [(validate.rules).map.values.double.gte = 0];
```
wbpcode, line 37: ditto.

wbpcode:
By the way, I think we may also need to add a validation to the `request_cost` to ensure every cost value is larger than or equal to 0?

wbpcode:
@adisuissa

xds/data/orca/v3/orca_load_report.proto
adisuissa: FWIW, to the best of my knowledge, the PGV validation rules are not actually used for validation at runtime, but just as a guideline to the implementers.

wbpcode: right, the validation only works if the call the validate method 🤔

adisuissa:
@efimki @markdroth any reservations about modifying the upper bound limits?

xds/data/orca/v3/orca_load_report.proto
adisuissa: FWIW, I this there should be a limit that is enforced in the code, regardless of whether the thresholds are explicitly stated here or not.
I would suggest that the code caps the range to something like [0, 1], just to prevent bad upstream actors from the ability to impact the load-balancing decision too much (either intentionally or unintentionally).

markdroth (commented):
> By the way, I think we may also need to add a validation to the request_cost to ensure every cost value is larger than or equal to 0?

Technically, *increasing* the strictness of a PGV annotation (as opposed to decreasing it) is considered a breaking API change.  The [API versioning doc](https://github.com/envoyproxy/envoy/blob/main/api/API_VERSIONING.md#backwards-compatibility) lists the following as a breaking change:

> Increasing the strictness of [protoc-gen-validate](https://github.com/bufbuild/protoc-gen-validate) annotations. Exceptions may be granted for scenarios in which these stricter conditions model behavior already implied structurally or by documentation.

In this particular case, though, if we know that Envoy doesn't actually use the PGV annotations for ORCA, then this change is probably safe.  As I mentioned in my other comment, gRPC does not use PGV at all, so this doesn't affect us.

PR #110, opened 2024-12-18
Add support for generating python proto code

coolkp:
Changes in the following PR
- Updated protoc version
- Generate go code only has new protoc version
- Update generate_go_protobuf -> generate_protobuf to add python code generation
- Generated Python code
- Package generated python code as pip package


python/setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import setup, find_packages
+
+setup(
+    name='xds',  
+    version='0.1.0', 
+    description='xds Protocol Buffer Messages', 
+    packages=find_packages(),
+    install_requires=[
+        'protobuf==5.29.1',  
+        'protoc-gen-validate',
```
adisuissa, line 10: Should this have a version?

python/setup.py
```diff
@@ -0,0 +1,12 @@
+from setuptools import setup, find_packages
```
adisuissa, line 1: What's the benefit of having both setup.py and pyproject.toml (in addition to the Bazel build that we have).
Seems we are adding more ways to generate python files out of protos and violate ODR while we are doing this. If this can be minimized, it would be best.

tools/generate_protobuf.py
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env python3
+
+from subprocess import check_output
+import glob
+import os
+import shutil
+
+def generate_go_protobufs(output,language):
```
adisuissa, line 8:
```suggestion
def generate_go_protobufs(output, language):
```

tools/generate_protobuf.py
```diff
@@ -0,0 +1,51 @@
...
+      output_file = shutil.copy(generated_file, output_dir)
+      os.chmod(output_file, 0o644)
+
+
+if __name__ == "__main__":
+  workspace = check_output(['bazel', 'info', 'workspace']).decode().strip()
+  output = os.path.join(workspace, 'go')
+  generate_go_protobufs(output, "go")
+  output = os.path.join(workspace, 'python')
+  generate_go_protobufs(output, "python")
```
adisuissa, line 51: The function name should probably be `generate_lang_files_from_protos`.

tools/generate_protobuf.py
```diff
@@ -0,0 +1,51 @@
...
+        'kind("go_proto_library", ...)',
+    ]).split()
+    output_dir = 'github.com/cncf/xds/python'
+  if language == "python":
+    protos = check_output([
+        'bazel',
+        'query',
+        'kind("py_proto_library", ...)',
+    ]).split()
+    output_dir = 'github.com/cncf/xds/go'
```
adisuissa, line 24: These should be passed as an input, rather than passing the `language`.
If possible, refactor the function to receive the parameters it needs and avoid the `if language ==` statements.

python/setup.py
```diff
@@ -0,0 +1,12 @@
...
+
+setup(
+    name='xds',
```
adisuissa, line 4: Will this impact the package name in pypi?
If so, it should probably be cncf-xds (just to avoid potential name conflicts with other packages)

python/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
adisuissa, line 1: Why was the MIT license chosen?
Should the license be kept in sync with https://github.com/cncf/xds/blob/main/LICENSE ?

adisuissa (commented):
Thanks!
In addition to the comments, a high-level question: are the `__init__.py` files auto generated? (I'm assume the bazel command adds these, but just want to make sure).

python/setup.py
coolkp: not sure what version we should have? what version is being used for go?

python/setup.py
coolkp: we are not thinking of publishing to pypi yet, since usage is going to be limited. We can just install the wheel directly.

python/setup.py
coolkp: consolidated into pyproject.toml. its the latest way of doing this

python/LICENSE
coolkp: removed it, its redundant since there is a parent license

tools/generate_protobuf.py
coolkp: Refactored

tools/generate_protobuf.py
coolkp: done

python/setup.py
adisuissa: It would be best to keep it in sync with https://github.com/cncf/xds/blob/cff3c89139a3e6a0d4fbddfd158ad895e9b30840/bazel/repository_locations.bzl#L15.

It is not great that we violate ODR here, but if simplicity in importing is desired (instead of using bazel), then I guess this must be the price.

tools/generate_lang_files_from_protos.py
```diff
@@ -0,0 +1,52 @@
...
+def generate_lang_files_from_protos(language: str):
+    output = os.path.join(workspace, language)
+    bazel_bin = check_output(["bazel", "info", "bazel-bin"]).decode().strip()
+
+    protos = check_output(
+            [
+                "bazel",
+                "query",
+                {"go":'kind("go_proto_library", ...)',
+                 "python":'kind("py_proto_library", ...)'}[language]
```
adisuissa, line 21: Although these are just 2 languages, this coding style can probably be improved.
I suggest having all the required definitions for GO/Python each in its own class, and have these as properties for each of the languages. The class will either be passed to `generate_lang_files_from_protos` or it will fetch the right one using a map at the beginning of the function.

adisuissa (commented):
Thanks, overall LGTM.
Just a suggestion on how to refactor the function to support multi-languages.

python/setup.py
coolkp: Btw this dependency is for the client, not for generating the code. VLLm client doesn't use bazel

coolkp: actually i removed this file itself and we don't need protoc-gen 😅

tools/generate_lang_files_from_protos.py
coolkp: Done

adisuissa (commented):
Thanks!
Overall LGTM, one comment about unifying requirements if possible.

python/xds.egg-info/PKG-INFO
```diff
@@ -0,0 +1,8 @@
+Metadata-Version: 2.1
+Name: xds
+Version: 0.1.0
+Summary: xds Protocol Buffer Messages
+Requires-Python: >=3.7
+License-File: LICENSE
+Requires-Dist: protobuf==5.29.1
+Requires-Dist: protoc-gen-validate
```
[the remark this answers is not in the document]
coolkp: this isn't needed, i removed it. was testing out the python package, this is output `pip install .`

tools/generate_lang_files_from_protos.py
```diff
@@ -0,0 +1,43 @@
...
+    ).split()
+    output_dir = f"github.com/cncf/xds/{language}"
+    check_output(["bazel", "build", "-c", "fastbuild"] + protos)
+    print(f"Found {len(protos)} bazel rules to generate code")
+    for rule in protos:
+        rule_dir = rule.decode()[2:].rsplit(":")[0]
+        input_dir = language_config.get_input_dir(bazel_bin, rule_dir)
+        input_files = glob.glob(os.path.join(input_dir, language_config.generated_file_pattern))    
+        output_dir = os.path.join(output, rule_dir)
+        print(f"Moving {len(input_files)} generated files from {input_dir} to output_dir {output_dir}")
```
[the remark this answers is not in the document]
adisuissa: Not sure why it got resolved.

PR #114, opened 2025-02-01
move go.mod to project root

### LINKED ISSUE #104: Gazelle: no such package '@com_github_cncf_xds_go//udpa/annotations': BUILD file not found in directory 'udpa/annotations'
We are using gazelle to generate bazel bindings to the golang code. The version `v0.0.0-20240905190251-b4127c9b8d78` fails with:

```
bazel run //:{redacted}
INFO: Repository org_golang_google_protobuf instantiated at:
...
INFO: repository @org_golang_google_protobuf' used the following cache hits instead of downloading the corresponding file.
 * Hash '39a8bbfadaa3e71f9d7741d67ee60d69db40422dc531708a777259e594d923e3' for https://mirror.bazel.build/github.com/protocolbuffers/protobuf-go/archive/refs/tags/v1.33.0.zip
If the definition of 'repository @org_golang_google_protobuf' was updated, verify that the hashes were also updated.
ERROR: /{redacted}/external/com_github_envoyproxy_go_control_plane/envoy/service/auth/v3/BUILD.bazel:3:11: no such package '@com_github_cncf_xds_go//udpa/annotations': BUILD file not found in directory 'udpa/annotations' of external repository @com_github_cncf_xds_go. Add a BUILD file to a directory to mark it as a package. and referenced by '@com_github_envoyproxy_go_control_plane//envoy/service/auth/v3:auth'
ERROR: Analysis of target '//:{redacted}' failed; build aborted:
INFO: Elapsed time: 2.685s
INFO: 0 processes.
ERROR: Build failed. Not running target
FAILED: Build did NOT complete successfully (87 packages loaded, 4389 targets configured)
```

The version `github.com/cncf/xds/go@v0.0.0-20240318125728-8a4994d93e50` works ok. It's regression in the newer versions.

My hypothesis about the root reason: @keith added in https://github.com/cncf/xds/pull/96 a [/go/BUILD](https://github.com/cncf/xds/blob/main/go/BUILD) file. 
The file makes gazelle consider this whole subtree as a proper BAZEL project... so suspends gazelle for generating the proper BUILD files in the sub tree... But they are missing there... i.e. there are no proper BUILD files, e.g. in: https://github.com/cncf/xds/tree/main/go/udpa/annotations. 

Can we make it either proper bazel project with gazelle generated BUILD targets... 
or rollback the BUILD file in the subtree at all ?

mmorel-35:
This moves go.mod to the root directory of the project and the go code follows this change to.

Closes #104


udpa/annotations/migrate.pb.go
adisuissa, line 1: Will it be possible to keep all those changes in the "go/" directory?
The reason is, I think it would be best to keep these files separated from the proto sources.
I'm not sure what causes these, but would like all the "go" related stuff that don't use bazel to be in their own directory.

adisuissa (commented):
Thanks for trying to fix this.
I'm not sure if it works for the people using go, but just wanted to highlight one thing about the locations of some files.

udpa/annotations/migrate.pb.go
adisuissa: I think this pattern should be discouraged.
IMHO having the go files so it is easier to use them is great, but it should be limited to a specific sub-dir that is different than where the "source files" (protos) are.

PR #115, opened 2025-02-08
add arbitrary expression to CelExpression

zirain:
this's suggested by @yanavlasov in https://github.com/envoyproxy/envoy/pull/38182#discussion_r1947017753

it's useful in envoy when there's no control plane.

cc @yanavlasov @htuch


xds/type/v3/cel.proto
```diff
@@ -47,6 +47,11 @@ message CelExpression {
   //
   // If set, takes precedence over ``cel_expr_parsed``.
   cel.expr.CheckedExpr cel_expr_checked = 4;
+
+  // Expression in abstract syntax tree (AST) form.
+  //
+  // If set, takes precedence over ``cel_expr_parsed`` and ``cel_expr_checked``.
+  string expression = 5 [(validate.rules).string = {min_len: 1}];
```
adisuissa, line 54: I think that the PGV validation makes this field required. If that's the case, I suggest to remove it.

xds/type/v3/cel.proto
```diff
@@ -47,6 +47,11 @@ message CelExpression {
   //
   // If set, takes precedence over ``cel_expr_parsed``.
   cel.expr.CheckedExpr cel_expr_checked = 4;
+
+  // Expression in abstract syntax tree (AST) form.
+  //
+  // If set, takes precedence over ``cel_expr_parsed`` and ``cel_expr_checked``.
```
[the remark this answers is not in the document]
zirain: updated

xds/type/v3/cel.proto
zirain: removed

xds/type/v3/cel.proto
```diff
@@ -47,6 +47,11 @@ message CelExpression {
   //
   // If set, takes precedence over ``cel_expr_parsed``.
   cel.expr.CheckedExpr cel_expr_checked = 4;
+
+  // Expression in abstract syntax tree (AST) form.
```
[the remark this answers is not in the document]
zirain: update comments

xds/type/v3/cel.proto
```diff
@@ -40,13 +40,20 @@ message CelExpression {
...
   cel.expr.ParsedExpr cel_expr_parsed = 3;
 
   // Parsed expression in abstract syntax tree (AST) form that has been successfully type checked.
   //
   // If set, takes precedence over ``cel_expr_parsed``.
+  // If ``expression`` is set, this field is not used.
   cel.expr.CheckedExpr cel_expr_checked = 4;
+
+  // Expression in abstract syntax tree (AST) form.
+  // This's useful when there's no control plane(e.g. XDS from file).
```
adisuissa, line 53: Sorry, I don't fully understand what's the realtionship between having a dynamic control-plane and having it in a file.
Can you shed some light here?

Also, are there any constraints for this field, or is there some format that needs to be used?
Will a generic string, e.g. "abc", make sense in this case?

zirain: Previous context is [here](https://github.com/envoyproxy/envoy/pull/38182#discussion_r1947017753), suggested by @yanavlasov.

I agree with that it's hard to send `cel.expr.CheckedExpr` when uses using file for xds instead of a control plane, or maybe I missed something.

zirain: it's a little painful to write something like following:
```yaml
    expression:
        parsed_expr:
          expr:
            id: 4
            call_expr:
              function: _!=_
              args:
              - id: 3
                select_expr:
                  field: id
                  operand:
                    id: 2
                    select_expr:
                      field: node
                      operand:
                        id: 1
                        ident_expr:
                          name: xds
              - id: 5
                const_expr:
                  string_value: "node_name"
```

yanavlasov: The use case is simple deployment or tests where it is easier to make Envoy parse the CEL expression, rather than having to specify the AST proto.

PR #116, opened 2025-02-10
Use protobuf 5.28.0

coolkp:
most clients use protobuf version greater than 5.28.0 e.g vllm, 5.29.1 causes compatibility issues
```bash
 google.protobuf.runtime_version.VersionError: 
Detected incompatible Protobuf Gencode/Runtime versions when loading 
xds/data/orca/v3/orca_load_report.proto: gencode 5.29.1 runtime 5.28.3. 
Runtime version cannot be older than the linked gencode version. See 
Protobuf version guarantees at https://protobuf.dev/support/cross-
version-runtime-guarantee.
```


adisuissa:
It seems counter productive to change the protobuf library version to an older version, just because there's a way to use pre-made proto generated files.
I wonder if it makes sense to search for an alternative, such as:
1) Update other projects to use the newer version.
2) Make the other projects build the proto versions to their end.
3) Remove the version from the generated files or set it to an older that is compatible with the current version.

PR #117, opened 2025-03-13
keep_matching in OnMatch to skip matched Matchers

bsurber:
Add a keep_matching field to OnMatch which indicates to callers that it should not enforce the matcher's action and instead continue looking for another match, if possible.
If a match continuation isn't possible or does not find a further match, on_no_match action will be used.

CC @markdroth


markdroth:
This LGTM, but I'll hold off on merging until we get consensus on https://github.com/envoyproxy/envoy/pull/38726.

bsurber:
> This LGTM, but I'll hold off on merging until we get consensus on [envoyproxy/envoy#38726](https://github.com/envoyproxy/envoy/pull/38726).

Now that we have consensus on the desired behavior, are we clear to submit this? It'll help clean up the CI + dependency management in the dependent PR.

markdroth:
Still need to address https://github.com/envoyproxy/envoy/pull/38726/files#r2014765850.

markdroth (approved):

markdroth:
Please fix DCO.

PR #118, opened 2025-03-26
cel: add raw string expression support

wbpcode:


xds/type/v3/cel.proto
```diff
@@ -47,6 +47,11 @@ message CelExpression {
   //
   // If set, takes precedence over ``cel_expr_parsed``.
   cel.expr.CheckedExpr cel_expr_checked = 4;
+
+  // Unparsed expression in string form.
+  //
+  // If set, takes precedence over ``cel_expr_parsed`` and ``cel_expr_checked``.
```
adisuissa, line 53: Please add a description or link to what is a valid string value that is expected here.

wbpcode: An example and the spec link is added.

adisuissa: Thanks!
I'm going to approve this for now, but it's not clear to me where is "request.headers" defined? Is there a "response.body"?
I would assume this is all written somewhere (doc), and there should be a link here to that doc.

adisuissa (approved):

xds/type/v3/cel.proto
wbpcode: Most CEL attributes are host-dependent. That's say, the example requires the host to support `request.headers` attributes. And you can see all attributes supported by envoy at here: https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/advanced/attributes

markdroth:
What's the motivation for this?

In general, I don't think we want to support sending raw CEL expressions to the data plane.  It makes more sense for the control plane to pre-compile the expression before sending it to the data plane, for both performance and validation reasons.

Also, note that gRPC C++ will basically never be able to support this, because CEL C++ depends on protobuf, and gRPC can't take a dependency on protobuf internally.

xds/type/v3/cel.proto
wbpcode: No sure because xds repo is also be used by other projects. So, I only put an example and spec link here.

PR #124, opened 2025-10-18
Migrate from deprecated protoc-gen-validate to protovalidate v1.0.0

mmorel-35:
#### Description

This is using protovalidate instead of protoc-gen-validate, also updates protobuf to 29.2 as it is transitively required by protovalidate


adisuissa (commented):
I'm a bit hesitant about these big changes as if there is some incompatibility it impact everything all at once (easy to find that it is broken, but many people won't be pleased).
Will it be possible to have the old and new protoc-gen-validators, and incrementally allow them in specific files?

I guess this may be more relevant to the Envoy code, as there's call to the validator there that may be different, but for the protos as well.

PR #127, opened 2025-11-01
Migrate from go_deps.from_file() to explicit go_deps.module() declarations

mmorel-35:
#### Description 

Fixes https://github.com/cncf/xds/issues/104
Closes https://github.com/cncf/xds/pull/112


mmorel-35:
https://github.com/envoyproxy/envoy/pull/41881 has passed with this version @adisuissa , when you have time to review?

mmorel-35:
The reason is that we have a dead end in https://github.com/cncf/xds/issues/104  and that is the easiest solution I could find that would remove BUILD in go directory

Issue #137, opened 2026-01-16
Mark xds/core CIDR as stable


fl0Lec:
The xds/core/v3/cidr.proto file still carries a work-in-progress annotation, but it has not changed in ~5 years and is widely used in production.

For example, Envoy uses this type in multiple APIs, including the RBAC filter:
https://github.com/envoyproxy/envoy/blob/09603b06b93c4f9278a096ad5beb2780eaf263fe/api/envoy/config/rbac/v3/rbac.proto#L273

Because of the WIP annotation, Envoy emits warnings at startup, even though this type is stable and relied upon by existing deployments.

Given its long-term stability and widespread usage, I propose removing the work-in-progress annotation from cidr.proto to avoid misleading users and unnecessary warnings.

PR #138, opened 2026-01-16
xds-core: mark CIDR as stable

fl0Lec:
removing the work-in-progress annotation to cidr.proto
the CidrRange seems to be used in a bunch of place notably in [Envoy RBAC](https://github.com/envoyproxy/envoy/blob/09603b06b93c4f9278a096ad5beb2780eaf263fe/api/envoy/config/rbac/v3/rbac.proto#L273).

issues for discussion: https://github.com/cncf/xds/issues/137


adisuissa (commented):
@kyessenov to confirm before we make it non-WIP.

kyessenov (approved):
Doesn't look particular controversial so accept this as done for now.

adisuissa (approved):
LGTM, thanks!

PR #145, opened 2026-05-18
Add ORCA specification

efimki:
Add a standalone ORCA specification under the top-level orca/ directory.

This specification is intended to be continuously updated as the format evolves.

Based on https://github.com/cncf/xds/pull/105 by @AndresGuedez


efimki:
@markdroth - please take a look when you have a chance. 
This PR should address your comments on #105.

